PR #13, opened 2021-06-11
Initial test. [Fixes #10]

### LINKED ISSUE #10: write initial correctness tests
Once you have added your initial code (#7) you should write tests that produce results that you know to be correct. 

Once you start refactoring your code, the tests will remain the same (at least the reference values ("fixtures") that they check against). This will provide you the confidence that your code changes do not introduce bugs. (In truth, you'll likely discover bugs in your old code and new bugs will creep in but still, your initial tests will be the foundation on which you're doing your work).

ojeda-e:
## Description
Initial test. Fixes [#10] This first version of the tests includes:
- Fixtures for universe, reference beads and cloud of Z points.
Test for lines (in `core.py`)
- line 76
- line 92
- line 99
- line 100
- line 101
- line 117-118
Test for functions (in `lib\mods.py`):
- line 391
- line 417

## Notes
- Some fixtures added for future work (leaflets and lipid types)
- The code currently uses `mdtraj`, therefore some fixtures are temporary and will be gone after refactoring.
- Added a new profile that contains only beads of reference to derive the surfaces (`GRO_PO4` in `test_mdakit_mmebcurv.py`).


ojeda-e:
Test are not failing here. 💪 
Now I really need to solve the documentation part (issue#5 , PR#8).

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,182 @@
...
+def Z_cloud(ref_beads):
+    n_cells, max_width = 10, 19
+    grid_count, grid_1, grid_2, z_ref = [np.zeros([n_cells, n_cells]) for i in range(4)]
+    factor = np.float32(n_cells / max_width)
+
+    for atom in ref_beads:
+        x, y, z = atom.position / 10
+
+        cell_l = int(abs(x) * factor)
+        cell_m = int(abs(y) * factor)
```
lilyminium, line 89: Hmm, I get nervous when I see coordinates being `abs`oluted. Just checking that you're sure that this puts them in the right cell? In my mind, an x value of -0.5 for a box length of 10 nm should put it in the last cell (9.5-10), not the first (0-0.5).

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,182 @@
...
+
+        try:
+            grid_1[cell_l, cell_m] += z
+            grid_2[cell_l, cell_m] += 1
+
+        except BaseException:
```
lilyminium, line 95: Which exception are you expecting here? Generally catching any possible exception is a bad idea in case something happens that you're really not expecting. Especially in a test, where expectations should be really specific.

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,182 @@
...
+            pass
+
+    for i, j in it.product(range(n_cells), range(n_cells)):
```
lilyminium, line 98: Is `it` itertools? It's a good idea to import it within this file, even if you've obtained it from `lib` with the `from lib.mods import *`

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,182 @@
...
+        if grid_2[i, j] > 0:
+            z_ref[i, j] += grid_1[i, j] / grid_2[i, j]
+            grid_count[i, j] += 1
```
lilyminium, line 101: Here is a nice place to take advantage of masking in numpy arrays. Does the below (untested) snippet accomplish the same thing for you?

```python
mask = grid_2 > 0

z_ref[mask] = grid_1[mask] / grid_2[mask]
```

However, I am a little confused as to the difference between `grid_count` and `grid_2`. Shouldn't they have the same values?

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,182 @@
...
+    factor = np.float32(n_cells / max_width)
+
+    for atom in ref_beads:
+        x, y, z = atom.position / 10
+
+        cell_l = int(abs(x) * factor)
+        cell_m = int(abs(y) * factor)
+
+        try:
+            grid_1[cell_l, cell_m] += z
```
lilyminium, line 92: Would you need to account for PBC with the z coordinate? With a box length of 10, the mean of a list of z-values [0.5, 9.5] should ideally be 0, not 5.

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,182 @@
...
+            pass
+
+    for i, j in it.product(range(n_cells), range(n_cells)):
+        if grid_2[i, j] > 0:
+            z_ref[i, j] += grid_1[i, j] / grid_2[i, j]
+            grid_count[i, j] += 1
+
+    for i, j in it.product(range(n_cells), range(n_cells)):
+        if grid_count[i, j] > 0:
+            z_ref[i, j] /= grid_count[i, j]
```
lilyminium, line 105: So just so I'm understanding this correctly, the z-ref coordinate is the sum of all the z-coordinates of the atoms in that cell, divided by the square of the count of atoms?

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,182 @@
...
+def leaflets():
+    return ['lower', 'upper']
+
+
+@pytest.fixture()
+def lipid_types():
+    return ['POPC', 'POPE']
+
 
 def test_mdakit_membcurv_imported():
```
lilyminium, line 126: Ehh you can probably delete this

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,182 @@
...
     """Sample test, will always pass so long as import statement worked"""
     assert "mdakit_membcurv" in sys.modules
+
+
+def test_output_folder():  # line 76, gone after refactoring
```
lilyminium, line 131: This test doesn't so much test your code as the `os` library. Is it necessary?

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,182 @@
...
+    max_width = universe.dimensions[0] * 0.1
+    assert max_width == 18.431013488769533
+
+
+def test_n_cells(universe):  # line 101
+    unit_width = 20
+    max_width = universe.dimensions[0] * 0.1
+    n_cells = math.ceil(max_width / unit_width * 10)
+    print(n_cells)
+    assert n_cells == 10.
```
lilyminium, line 158: These tests are very fine-grained. `test_box_size` is mostly checking your assumptions of how MDAnalysis reads dimensions, which is fine, but `test_max_width` seems to be testing that division by 10 works as expected -- I think that is probably unnecessary.

In general the tests in this repository should be for *your* code. A good rule of thumb is that each test should test the output of a function that you wrote. So if you had a function called `get_max_width`, you could test that, but IMO it's out of scope to test `MDAnalysis` and `math` functionality.

lilyminium (changes requested):
Thanks for starting up the tests! I had a few questions about the Z-points and some notes about tests :)

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: This can trigger a Lon conversation about the method I implemented, but let's try :) 
I used absolute values because of two main reasons. 1) I am dismissing those few atoms that have -2.x and lower (is never higher than that, otherwise that would be reasonable.) 
With that range, when negative, the bead will be very likely mapped to first or second unit cell. 
In previous attempts I tried to generate a cells for those rare cases, but it messed up the grid because then it will introduce lots of `np.nans`, and that will drive the values of the surface pretty badly. 
2) I am not considering PBD conditions in this method. 

Despite these, I am pretty sure of the `abs` there.

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: Actually that was a autopep8 addition. My initial code had except pass.

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: I can start cleaning up this a bit. 😅

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: Yeah, I left it because I may use it in a future version of the test. I can't get rid of it too.

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: Thanks. 
Maybe I need more directions at this point. I remembered I asked @orbeckst about what part of the code should I test to start, considering I have a `main.py` and `mods.py`. I tried to test main, with the exception of the `parse_arg` part and some other redundant functions like `curvature` that I wrote just to made my life easier when I was using it for me. 
I thought what I had in this version of the tests was checking some lines in `main.py` and some other is `mods.py`. Not all of them, because there are parts of the code that will be gone in the refactoring. 
Maybe mentors can help giving me some lines to test? ☺️

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: `z_ref` is the sum of all the `z` coordinates in that cell, divided by the total number of atoms that populated that cell (`grid_count` stores that number).

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: Hmmm... interesting. I guess I could, but that system doesn't make sense in my mind. In that type of setups I'd say the simulation box is too small and the interactions of the images will affect the dynamic of the atoms in the simulation box itself. Maybe I am missing something here?

lilyminium: I'm thinking of the scenario that the membrane is on the "bottom" of the box on the z-axis instead of the middle, i.e. around the origin -- maybe rare but very possible.

mdakit_membcurv/tests/test_mdakit_membcurv.py
lilyminium: But I think you divide it by `grid_2` already, which also stores the atom count?

mdakit_membcurv/tests/test_mdakit_membcurv.py
lilyminium: It's just testing that it's successfully imported, which I think is usually a given for a package -- I suppose it could catch errors in installation.

mdakit_membcurv/tests/test_mdakit_membcurv.py
lilyminium: Maybe add at least one test for every function that you think you will need in some form in the MDAnalysis refactor? Even if the form of the function changes when refactoring, we would want to check that the returned output for the given input does not.

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: Holy. Yeah, I wasn't prepared for that type of escenarios. In that case I'll have to add PBC on `z`. Thanks :)

mdakit_membcurv/tests/test_mdakit_membcurv.py
lilyminium: > With that range, when negative, the bead will be very likely mapped to first or second unit cell.

Could you not subtract the coordinates from the box size to support PBC? I think the MDAnalysis refactor will very likely include PBC so it's a good idea to start thinking about it now.

mdakit_membcurv/tests/test_mdakit_membcurv.py
lilyminium: `except` is also quite general, maybe more so than `except BaseException`. As this is a test where we are defining our expectations for the code, we just want to be as specific as possible :)

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: I see your point now. In this test I am not iterating over frames. I am actually extracting the surface and calculating `H` and `K` from the file with PO4 beads only as in getting one frame from the gro file. Very likely I missed or added something but can't see it now. I'll review it. Thanks.

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: I think is better if I delete it for now since I am not using it. Thanks.

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: Ok, thanks.

mdakit_membcurv/core.py
```diff
@@ -5,7 +5,7 @@
 Handles the primary functions
 """
 
-from mods import *
+from .lib.mods import *
```
orbeckst, line 8: I would avoid blanket `from x import *` imports but be explicit about which functions you actually need here. It makes the code a lot easier to debug and to read. Furthermore, if you have two blanket imports, _nobody_ (including future you) will likely have an idea where the functions really came from.

(I am not even a fan of `from x import y` and prefer to use `import y` and then use `y.x` to make clear where it comes from, or use a variation with submodule imports and aliasing like `from matplotlib import pyplot as plt`.)

mdakit_membcurv/tests/datafiles.py
```diff
@@ -0,0 +1,23 @@
+
+
+"""
+Location of datafiles for Membrane Curvature unit tests
+=======================================================
+
+MD simulations files stored in ``\\data`` sub-directory.
```
orbeckst, line 7: I'd write it as
```reST
MD simulations files stored in ``data`` sub-directory.
```
and not start with a leading directory separator. (Also, I _think_ convention is to use Unix forward slashes as directory separators in documentation... but I might be biased.)

mdakit_membcurv/tests/datafiles.py
```diff
@@ -0,0 +1,23 @@
...
+
+    from mdakit_membcurv.datafiles import *
+"""
+
+__all__ = [
+    "GRO_",  # Gromacs file of POPC POPE CHOL membrane
+    "XTC_"  # Gromacs trajectory of 10 frames.
```
orbeckst, line 14: Maybe be more explicit and name them `GRO_MIXED_MEMBRANE` and `XTC_MIXED_MEMBRANE` instead of a trailing underscore?

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,274 @@
 """
 
 # Import package, test suite, and other packages as needed
-import mdakit_membcurv
+import pickle
+
+from numpy.testing._private.utils import assert_array_almost_equal
```
orbeckst, line 8: Does
```python
from numpy.testing import assert_array_almost_equal
```
not work? It would be simpler and not rely on an import that seems to not be guaranteed to exist in future versions. I'd avoid using anything starting with underscores as it is generally not part of the published API.

Btw, we typically use `assert_almost_equal`.

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,274 @@
...
 import pytest
 import sys
+import math
+import os
+import mdtraj as md  # This will be gone after refactoring
+import itertools as it
+from ..lib.mods import *
```
orbeckst, line 15: explicit imports

mdakit_membcurv/tests/test_mdakit_membcurv.py
orbeckst: ... as you do below, so I think the above can just go.

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,274 @@
...
+
+@pytest.fixture()
+def ref_beads():
+    u = mda.Universe(GRO_PO4)
+    ref_beads = u.atoms[0:455]
+    return ref_beads
+
+
+@pytest.fixture()
+def Z_cloud(ref_beads):
```
orbeckst, line 165: Is this code reproducing some functionality in the main code? You are doing a fair amount of calculation to set up a reference value. Can you add some commentary here to make clear what happens? 

This is even more important in the light of https://github.com/MDAnalysis/membrane-curvature/pull/13/files#r650475495 if there are concerns how PBC are treated here. If you are generated an incorrect reference and then test your code against it, you will ensure that your code is incorrect. I understand that at this stage you want to test that your code produces the values that it has been producing in the past. If during writing the tests you find potential bugs in your code then you have two options (in my opinion)

1. write correct test and [mark them XFAIL](https://docs.pytest.org/en/6.2.x/skipping.html#xfail-mark-test-functions-as-expected-to-fail) (i.e., you know your code will currently not produce the correct result) and fix later
2. write correct tests and fix your code now

If things are not clear cut then at least add a comment in the tests and in your main code noting that this is something that needs to be looked into.

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,274 @@
...
+    assert n_cells == 10.
+
+
+def test_dict_to_pickle():  # line 117-118
+    name = 'test_pickle_output'
+    dict_ = {'A': 1, 'B': 2, 'C': 3}
+    dict2pickle(name, dict_)
+    unpickled = pickle.load(open(name + '.pickle', 'rb'))
+    assert dict_ == unpickled
+    os.remove(name + '.pickle')
```
orbeckst, line 232: Tests that create files should use the [`tmpdir` fixture](https://docs.pytest.org/en/6.2.x/reference.html?highlight=tmpdir#std-fixture-tmpdir)
```python
def test(... tmpdir):
   with tmpdir.cwd():
        # create files etc
        do_test
        assert something
```
When the `with` block finishes, `tmpdir` and everything inside will be autodeleted.

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,274 @@
...
+
+
+def test_md_traj():  # line 108, gone after refactoring
+    traj = md.load(XTC_, top=GRO_)
+    assert traj.n_frames == 11
```
orbeckst, line 237: Not needed, just checks that your trajectory has the length that you think it should have. This does not check _your_ code.

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,274 @@
...
+
+
+def test_gaussian_curvature(Z_cloud):  # line 391 of mods
```
orbeckst, line 240: This is an important test as it checks _your_ function.

orbeckst (commented):
Good progress, please see comments inline.

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: I removed the exception in the test file for now, but yes, I'll make sure of checking this part in the refactoring. :)

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: It was my bad for putting it here. I'll try to explain. Initially the code (in `main.py`) has two grids to count how many beads lie in a given unit cell, one that is associated to the total number of beads over the full iteration of the n frames (`grid_count`), and one associated to the beads of each lipid type per frame (`grid_2`). 
In the `Z_cloud` fixture, I removed the iteration over frames since I am extracting the cloud of Z points out of the `.gro` file containing PO4 beads only.

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: It partially does, yes. It is a short calculation to map the `(x,y)` to `[i,j]` of the beads in the grofile that contains `PO4` beads only. The subsequent calculation of membrane curvature uses the grid I am returning here, which is a `np.array`, indexed according to number of cells in the grid, and which contains the average `z` values of the beads that populate such unit cell. I will add comment on the steps. It's the same way it is calculated in `mods.py`, [line 291-303](https://github.com/MDAnalysis/membrane-curvature/blob/main/mdakit_membcurv/lib/mods.py).
My original code doesn't consider PBC conditions of any kind, and I'm aware that's a potential risk. The calculation of membrane curvature in this approach has an exception for the atoms of coordinates that when mapped to a grid, they fall in an `[i,j]` pair bigger than the initial size of the box.

mdakit_membcurv/core.py
ojeda-e: Thanks for this. I'll leave it as absolute import in the current test as `from mods import y` bit I will keep in mind the change as you suggested in the refactoring process.

mdakit_membcurv/tests/datafiles.py
ojeda-e: Thanks, changed to a more explicit name.

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: Thanks, I managed to miss this comment. Fixture added in last commit.

ojeda-e:
Hi @lilyminium  and @orbeckst.
I submitted the requested changes and replied to some of the comments you made in the previous PR. Thanks.

mdakit_membcurv/tests/test_mdakit_membcurv.py
lilyminium: I agree with @orbeckst that I am worried this is checking for incorrect functionality. Instead of having this function and loading ref_beads from a file, could you simply define Z_cloud as the z_ref array?

The simplest way to do this would be to run the code yourself and copy the coordinates into here, then we can work up to making sure that the generation of z_ref is correct.

lilyminium (changes requested):

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,290 @@
...
+def mdtraj_po4():
+    # trajectory PO4 beads only imported using mdtraj, gone after refactoring
+    mdtraj = md.load(XTC_PO4, top=GRO_PO4)
+    return mdtraj
+
+
+@pytest.fixture(scope="session")
+def output(tmpdir_factory):
+    tmp = tmpdir_factory.mktemp("output")
+    return tmp
```
lilyminium, line 213: Instead of this, you could use `tmpdir` directly in your function.

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,290 @@
...
+    assert n_cells == 10.
+
+
+def test_dict_to_pickle(output):  # line 117-118
+    name = 'test_pickle_output'
+    dict_ = {'A': 1, 'B': 2, 'C': 3}
+    with output.as_cwd():
+        dict2pickle(name, dict_)
+        unpickled = pickle.load(open(name + '.pickle', 'rb'))
+        assert dict_ == unpickled
```
lilyminium, line 253:
```suggestion
def test_dict_to_pickle(tmpdir):  # line 117-118
    name = 'test_pickle_output'
    dict_ = {'A': 1, 'B': 2, 'C': 3}
    with tmpdir.as_cwd():
        dict2pickle(name, dict_)
        unpickled = pickle.load(open(name + '.pickle', 'rb'))
        assert dict_ == unpickled
```

`tmpdir` is a fixture provided by pytest, so you don't have to make it yourself -- you could just use it directly here.

mdakit_membcurv/tests/test_mdakit_membcurv.py
lilyminium: You could then also write a test for the lines of code that generate `z_ref` :)

mdakit_membcurv/tests/test_mdakit_membcurv.py
```diff
@@ -3,10 +3,290 @@
...
+    top = md.load(GRO_MEMBRANE_PROTEIN).topology
+    assert top.n_atoms == 18891
+
+
+def test_box_size(universe):  # line 99
+    box_size = universe.dimensions[0]
+    assert_almost_equal(box_size, 184.31013, decimal=5)
+
+
+def test_n_cells(universe):  # line 101
```
lilyminium, line 239: `test_box_size, test_dict_to_pickle, test_md_topology` don't test *your* code so they should get removed

mdakit_membcurv/tests/test_mdakit_membcurv.py
ojeda-e: Thanks, I didn't realize this. Changed now. It's gone now, but a good exercise.

PR #25, opened 2021-06-18
Docs build

### LINKED ISSUE #24: Docs not building after renaming.
`Import membrane_curvature` (Line 16) seems to be the problem. 
Docs build otherwise. 

Any suggestion @lilyminium on how to solve this issue?

ojeda-e:
Fixes #24

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,7 +13,7 @@
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 
 # Incase the project was not installed
-import membrane_curvature
+# import membrane_curvature
 import os
 import sys
 sys.path.insert(0, os.path.abspath('..'))


ojeda-e:
I will merge this PR since it is a very simple change @lilyminium @IAlibay

IAlibay:
It's not immediately clear to me as to why the import is failing. I think I might be missing something?

ojeda-e:
Thanks for your answer @IAlibay.
It's is not clear to me either. I can't understand why the change here included fixes the issue. 
Even worse considering the test in the CI passes. 

Maybe we can leave this issue open?

IAlibay:
I think I'll defer to @lilyminium here, but if it works in CI but not locally (am I correct in understanding this is the issue you encountered?) I would suspect a conflict in your local environment. I assume you can import that module locally? (outside this test)

ojeda-e:
> I think I'll defer to @lilyminium here, but if it works in CI but not locally (am I correct in understanding this is the issue you encountered?) 

Yes, correct.
And I think you are right @IAlibay and the problem is in my virtual env. Let's wait for @lilyminium's opinion on this.
Thanks ☺️

Issue #27, opened 2021-06-18
Write test `def_all_beads`


ojeda-e:
Write test for `def_all_beads` in `mods.py` to cover lines 233-241.

This issue will help to close #17

Issue #17, opened 2021-06-13
Create more specific issues and add to suitable milestones


lilyminium:
As #13 notes, while it introduces some initial tests for the final result of each curvature function, additional and more specific intermediate tests (e.g. for getting the z-coordinates from a particular trajectory) will be added. In order to keep an eye on progress with the automated project tracker (https://github.com/MDAnalysis/membrane-curvature/projects/1), perhaps you could create some more issues for what you plan to do in the coming weeks, @ojeda-e?

For example, instead of "write tests for MDTraj code", an issue could be "write tests for the core_fast_* functions" (and add it to the [MDTraj](https://github.com/MDAnalysis/membrane-curvature/milestone/2) milestone). You could also look further ahead and start planning the structure of the MDAnalysis re-factor -- which functions you want to do first, etc.

~I *think* the project will automatically track created issues and PR's, so the only manual work is adding labels and/or milestones.~ Edit: sorry, you will have to manually add issues and PR's to the project tracker and milestones. This is quite a lot of project management so let us know if it gets too redundant and you prefer one or the other -- I am new to GitHub project management too!

I do like milestones because they have deadlines, so you can see how you're going in a particular area time-wise. They're usually overarching goals that have many issues and PR's attached (e.g. MDAnalysis uses them for [releases](https://github.com/MDAnalysis/mdanalysis/milestone/16)). I would guess that "MDAnalysis refactor" would be the next big one. The milestones tab can be a bit hard to find so I will just it here: https://github.com/MDAnalysis/membrane-curvature/milestones?direction=asc&sort=due_date

ojeda-e:
I have created more specific issues, all of them assigned to their respective milestones and properly labeled. At least to what I can see in the short term. I could add long term ones, but will be very likely changed/replace for others.
Should I keep this issue open @lilyminium?

ojeda-e:
After deleting unnecessary/unused lines in PR #29, coverage was increased by 15.92%.

lilyminium:
Thanks @ojeda-e for opening the new issues, with these and the plans you showed us in your presentation I think this issue is good to close :-)

ojeda-e:
Thanks @lilyminium :)

PR #30, opened 2021-06-18
test core_fast added [Issue #28]

ojeda-e:
Fixes issue #28 

## Changes
- [x] test_core_fast added.
- [x] removed unused argument in the `core_fast` function in `core.py`.
- [x] removed unnecesary tests functions.

This PR helps to solve #17

PR #31, opened 2021-06-19
test core_def_all_beads added. [Issue #27]

ojeda-e:
Fixes issue #27 

Changes:

- [x] test_def_all_beads added.
- [x] added import `def_all_beads` in `core.py` 

This PR helps to solve #17


membrane_curvature/core.py
```diff
@@ -18,6 +18,7 @@
 
 sys.path.append('lib/')
```
lilyminium, line 19: Can you see if you can remove this line? It shouldn't be necessary; if something breaks, you may need to `from .lib import my_func` before calling `my_func`.

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -5,13 +5,15 @@
 # Import package, test suite, and other packages as needed
 import pickle
 
+from numpy.testing._private.utils import assert_array_equal
```
lilyminium, line 8: In general you should avoid using private paths (typically prefixed with an underscore), as they may change in the future. Can you `from numpy.testing import assert_array_equal)?

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -239,3 +278,14 @@ def test_curvature(md_ref_beads, mdtraj_po4):
...
+
+
+def test_def_all_beads():
+    lipid_types = ['POPC', 'POPE']
+    lfs = ["upper", "lower"]
+    head_list = [4286, 11531, 18891]
+    topology = md.load(GRO_MEMBRANE_PROTEIN).topology
+    beads_test = def_all_beads(lipid_types, lfs, head_list, topology)
+    for lt in lipid_types:
+        for bead, bead_t in zip(MEMBRANE_CURVATURE_DATA['beads']["upper"][lt], beads_test["upper"][lt]):
```
lilyminium, line 290: What about the lower beads?

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -239,3 +278,14 @@ def test_curvature(md_ref_beads, mdtraj_po4):
...
+            assert_almost_equal(int(bead), int(bead_t))
```
lilyminium, line 291: As this is an integer, this should be fully equal instead of almost equal (which is usually used for floats). You should also know that both values are already integers, right? Could you `assert bead == bead_t`?

lilyminium (changes requested):
Thanks @ojeda-e for adding this test. I assume it's a regression test to test former behaviour. Could you please add a smaller example that's more easily human-verifiable (e.g. 10 lipids in each leaflet)? Looking at this test I have no idea what `def_all_beads` is for, nor how the head_list is meant to work -- it looks as though you've only told the function that 3 lipid indices belong to different leaflets, and yet I see that there are many beads in the MEMBRANE_CURVATURE_DATA reference. What happens if your beads are all mixed up, like lipid 1 and 3 are in the upper leaflet and lipid 2 is in the lower one? If your code doesn't currently allow for that, it's fine for the test to fail (mark it with [pytest.mark.xfail](https://docs.pytest.org/en/6.2.x/skipping.html) and give a reason).

membrane_curvature/tests/test_mdakit_membcurv.py
ojeda-e: Sorry, I thought I added that change already in a previous PR. That change wasn't saved.

membrane_curvature/tests/test_mdakit_membcurv.py
ojeda-e: >Could you please add a smaller example that's more easily human-verifiable (e.g. 10 lipids in each leaflet)? 
Should I add a new test (smaller number of beads), or is it better to replace the current one?

lilyminium: A new test in addition to this would be the best of both worlds ^_^

ojeda-e:
Thanks for the review @lilyminium. Here some answers (and a question) :)

> it looks as though you've only told the function that 3 lipid indices belong to different leaflets, and yet I see that there are many beads in the MEMBRANE_CURVATURE_DATA reference. 

To clarify, the indexes for leaflets were initially provided by the user with parse_args, `-ii`, and `-io`. For the test, the associated `grofile` has the indexes
outer leaflet: `range(4286, 11531)`
inner leaflet: `range(11531, 18891)`

> What happens if your beads are all mixed up, like lipid 1 and 3 are in the upper leaflet and lipid 2 is in the lower one? If your code doesn't currently allow for that, it's fine for the test to fail (mark it with [pytest.mark.xfail](https://docs.pytest.org/en/6.2.x/skipping.html) and give a reason).

It will probably assign the wrong indexes to each leaflet. It'll fail for sure. But now a question: in the structure of a `grofile`, the lipids are always "organized", aren't they? I mean. The way indexes are assigned to elements (atoms or beads) in a `grofile` is always strictly increasing. 
If that's true, even in cases where flip-flop occurs (trajectory), the `grofile` won't mix up the indexes per leaflet. Please correct me if I am wrong.

lilyminium:
> It will probably assign the wrong indexes to each leaflet. It'll fail for sure. But now a question: in the structure of a grofile, the lipids are always "organized", aren't they? I mean. The way indexes are assigned to elements (atoms or beads) in a grofile is always strictly increasing.

It's increasing, but only in terms of "this is the 58th lipid, this is the 59th lipid". It's pretty arbitrary where the lipid is in 3D space and does not correspond with the order in the gro file. The reason most membranes are divided by leaflet in the grofile is because that's how most humans think, and mostly humans put the grofile together.

> If that's true, even in cases where flip-flop occurs (trajectory), the grofile won't mix up the indexes per leaflet. Please correct me if I am wrong.

The order of the grofile also remains static; only the positions change. That means that even if cholesterol flips to the other leaflet, the index won't change. So you could have lower-leaflet cholesterol mixed in with the other upper leaflet lipids.

ojeda-e:
> The order of the grofile also remains static; only the positions change. That means that even if cholesterol flips to the other leaflet, the index won't change. So you could have lower-leaflet cholesterol mixed in with the other upper leaflet lipids.

Sorry I am not communicating correctly. The code reads only once the `grofile`, before iterating any frame. `def_all_beads` is not called in every frame, otherwise, it would be very expensive to run and would be affected by the 3D instant coordinates it adopts for every `dt`. The `grofile` is static, and for the same reason by reading it once, before iterating over any frame, indexes are ordered. Hence, if flips occur, that won't change the indexes assign per leaflet.

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -163,24 +163,121 @@
...
+               16008, 16020, 16032, 16044, 16056, 16068, 16080, 16092, 16104, 16116, 16128,
+               16140, 16152, 16164, 16176, 16188, 16200, 16212, 16224, 16236, 16248, 16260,
+               16272, 16284, 16296, 16308, 16320, 16332, 16344, 16356, 16368, 16380, 16392,
+               16404, 16416, 16428, 16440, 16452, 16464, 16476, 16488, 16500, 16512, 16524,
+               16536, 16548, 16560, 16572, 16584, 16596, 16608, 16620, 16632, 16644, 16656,
+               16668, 16680, 16692, 16704, 16716, 16728, 16740, 16752, 16764, 16776, 16788,
+               16800, 16812, 16824, 16836, 16848, 16860, 16872, 16884, 16896, 16908, 16920,
+               16932, 16944, 16956, 16968, 16980, 16992, 17004, 17016, 17028, 17040]}},
+
+    'small': {'lower': {'POPC': [5, 6, 7], 'POPE': [8, 9]}, 'upper': {'POPC': [0, 1, 4], 'POPE': [2, 3]}},
```
lilyminium, line 253: The upper POPC molecules of [0, 1, 4] have the coordinates:
>  1923POPC   PO4    1   1.361   0.630  16.101
>  2054POPC   PO4    2   0.704   2.229  15.981
>  2622POPC   PO4    5   1.451   0.382  *11.752*

That last POPC looks much more similar to the lower POPC with the coordinates:
>  2702POPC   PO4    6   0.081   0.610  11.835
>  2708POPC   PO4    7   0.678   2.233  11.825
>  2713POPC   PO4    8   0.988   0.554  11.838

Are you sure this is the right answer?

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -163,24 +163,121 @@
...
+
+    'med': {'lower': {'POPC': [12, 13, 14, 15, 16],
+                      'POPE': [17, 18, 19, 20, 21, 22, 23, 24]},
+            'upper': {'POPC': [0, 1, 2, 3, 4, 5, 6, 7, 11],
+                      'POPE': [8, 9, 10]}},
```
lilyminium, line 258: The z-positions of the upper POPC are: `[16.101, 16.251, 16.095, 15.981, 16.518, 16.049, 15.95 , 16.268, 11.752]`. That last POPC also seems quite low.

I haven't checked this for the big system but could you please check that the results you're getting are correct?

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -163,24 +163,121 @@
...
+
+    'big': {'lower': {'POPC': [25, 26, 27, 28, 29, 30, 31, 32, 33, 34],
+                      'POPE': [35, 36, 37, 38, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49]},
+            'upper': {'POPC': [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 23, 24],
+                      'POPE': [15, 16, 17, 18, 19, 20, 21, 22]}}
 }
 
 
-@pytest.fixture()
+@ pytest.fixture()
```
lilyminium, line 267: I don't think it's normally conventional to put a space after the @?

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -239,3 +336,59 @@ def test_curvature(md_ref_beads, mdtraj_po4):
...
+
+def test_def_all_beads_membrane_protein():
+    lipid_types = ['POPC', 'POPE']
+    lfs = ["upper", "lower"]
+    head_list = [4286, 11531, 18891]
+    topology = md.load(GRO_MEMBRANE_PROTEIN).topology
+    beads_test = def_all_beads(lipid_types, lfs, head_list, topology)
+    for lt, lf in zip(lipid_types, lfs):
+        for bead, bead_t in zip(MEMBRANE_CURVATURE_DATA['membrane_protein'][lf][lt], beads_test[lf][lt]):
+            assert int(bead) == int(bead_t)
```
lilyminium, line 349: Do you need these `int()` calls? Aren't they already integers?

lilyminium (changes requested):

lilyminium:
In addition, instead of using selections from the GRO file, I was hoping you could drum up some fake coordinates as outlined in https://github.com/MDAnalysis/membrane-curvature/pull/30#pullrequestreview-687924575 so that you could use the same system for checking the x-y gridding as well.

lilyminium:
I think my comment on this got lost in the review -- thanks for adding the smaller files, they made it easier to go through and verify that we are or aren't getting the results we want :)

membrane_curvature/tests/test_mdakit_membcurv.py
ojeda-e: Interesting. I just noticed it's the pep8 plugin (in VS) adds those spaces. Deleted in the last commit.

ojeda-e:
Thanks for your review, @lilyminium 
I think at this point is better to reduce the test to only one small system, which is here included as `test_po4_small.gro` and its respective trajectory file `test_po4_small.xtc` which includes 2 frames. The simulation box has dimensions of `x_dim, y_dim = 3, 3`. Therefore, the grid will be considered for this test as a 3x3 grid. 
The medium and big systems were temporarily removed.

Given the functions I should test, both gro and xtc files are needed. Fixing issue #32 will allow testing functions with dummy coordinates instead of `grofiles`.

In what respect to this PR, the small gro file provided includes 9 beads, all of them POPC lipid type, to simplify. Also to simplify, all beads have the same `z` value: `z=15.000` in the upper leaflet, and `z=12.000` in the lower leaflet. Since I am interested in checking `z_avg` and the mapping, realistic values are not relevant at this point. 
 
There are 4 beads in the upper leaflet, beads [(0) to (3)], and 5 beads in the lower leaflet, beads [(4) to (8)]. Or as provided, as in the pytest.parametrized, it can be read as: `ref_beads = {'small': {'upper': {'POPC': [0, 1, 2, 3]}, 'lower': {'POPC': [4, 5, 6, 7, 8]}}}`

The 9 lipids are distributed as shown in the figure below

```
Frame 1

^ o ___ o ___ o __ |
| |     |     |    |
| (6)__(7)___(8)___|
y |     |     |    |
| (3)__(4)___(5)___|
| |     |     |    |
v (0)__(1)___(2)___|
  0.    1.    2.   3.
  <------ x ------->

Frame 2

^  o ___ o ___ o ___ |
|  | (6) | (7) | (8) |
|  o ___ o ___ o ___ |
y  | (3) | (4) | (5) |
|  o ___ o ___ o ___ |
|  | (0) | (1) | (2) |
v  o ___ o ___ o ___ |
  0.    1.    2.   3.
   <------- x ------->
```
Each bead moves 0.5 in both (x,y) directions, between frame 1 and 2. 
Each bead is mapped to its respective unit cell, according to its `(x,y)` coordinates.

==== In  upper leaflet:
Frame No. 0
Bead  0 mapped to: [ 0 0 ]  with coords: 0.0 0.0 15.0
Bead  1 mapped to: [ 1 0 ]  with coords: 1.0 0.0 15.0
Bead  2 mapped to: [ 2 0 ]  with coords: 2.0 0.0 15.0
Bead  3 mapped to: [ 0 1 ]  with coords: 0.0 1.0 15.0
Frame No. 1
Bead  0 mapped to: [ 0 0 ]  with coords: 0.5 0.5 15.0
Bead  1 mapped to: [ 1 0 ]  with coords: 1.5 0.5 15.0
Bead  2 mapped to: [ 2 0 ]  with coords: 2.5 0.5 15.0
Bead  3 mapped to: [ 0 1 ]  with coords: 0.5 1.5 15.0

==== In  lower leaflet:
Frame No. 0
Bead  4 mapped to: [ 1 1 ]  with coords: 1.0 1.0 12.0
Bead  5 mapped to: [ 2 1 ]  with coords: 2.0 1.0 12.0
Bead  6 mapped to: [ 0 2 ]  with coords: 0.0 2.0 12.0
Bead  7 mapped to: [ 1 2 ]  with coords: 1.0 2.0 12.0
Bead  8 mapped to: [ 2 2 ]  with coords: 2.0 2.0 12.0
Frame No. 1
Bead  4 mapped to: [ 1 1 ]  with coords: 1.5 1.5 12.0
Bead  5 mapped to: [ 2 1 ]  with coords: 2.5 1.5 12.0
Bead  6 mapped to: [ 0 2 ]  with coords: 0.5 2.5 12.0
Bead  7 mapped to: [ 1 2 ]  with coords: 1.5 2.5 12.0
Bead  8 mapped to: [ 2 2 ]  with coords: 2.5 2.5 12.0

As a result, the first grid will be populated 

Frame 1
- Upper Leaflet
```
o ______ o _____ o _______ |
| np.nan | np.nan | np.nan |
o ______ o _____ o _______ |
|   (3)  | np.nan | np.nan |
o _______o ______ o ______ |
|   (0)  |   (1)  |   (2)  |
o ______ o ______ o ______ |
```
-Lower Leaflet
```
o ______ o _____ o _______ |
|   (6)  |  (7)   |   (8)  |
o ______ o _____ o _______ |
| np.nan |  (5)   |   (6)  |
o _______o ______ o ______ |
| np.nan | np.nan | np.nan |
o ______ o ______ o ______ |
```

The diagrams above are equivalent to 
```
z = {'upper': array([[15., 15., nan], [15., nan, nan], [15., nan, nan]]), 
     'lower': array([[nan, nan, 12.], [nan, 12., 12.], [nan, 12., 12.]])}
```

With these values from the small system, I added tests for the functions `core_fast`, `core_fast_leaflet`, and `test_all_beads` in mods.py.
Although I could have parametrized fixtures, I decided not to because this test will dramatically change once `MDtraj` is replaced by `MDAnalysis`, which is the next step.  
I hope these tests are simpler to follow, then I can focus on refactoring.
Thanks.

ojeda-e:
@IAlibay, maybe before creating universes from arrays and write and re-load, could you please take a look here?

membrane_curvature/tests/test_mdakit_membcurv.py
orbeckst: It might think (wrongly) that `@` is the binary matrix multiplication operator??

In MDAnalysis we typically look how we wrote existing code and then try to match the existing code, unless it appears to go against our [style guide](https://userguide.mdanalysis.org/dev/contributing_code.html?highlight=style%20guide#code-formatting-in-python)

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -174,10 +268,23 @@ def universe():
 
 
 @pytest.fixture()
-def mdtraj_po4():
+def leaflets():
+    return ["lower", "upper"]
```
lilyminium, line 272: This isn't *wrong*, but you could also just have the line below somewhere high up in the file, or pass in the leaflet labels with `pytest.mark.parametrize`.

```
LEAFLETS = ["lower", "upper"]
```

Fixtures are initialized and destroyed for each call of a test function (depending on the scope of the fixture) in a clean way to make sure that the input is always the same. That's not really necessary for a list of strings.

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -174,10 +268,23 @@ def universe():
...
+
+
+@pytest.fixture()
+def topol():
+    top = {'small': md.load(GRO_PO4_SMALL).topology,
+           'all': md.load(GRO_PO4).topology}
```
lilyminium, line 278: This on the other hand is a good use of a fixture. If one test modifies the GRO_PO4 MDTraj object, we don't want that to affect another test.

lilyminium (approved):
Thanks for making the changes @ojeda-e! I have some small comments but I'll approve because they are quite small.

ojeda-e:
Thanks for the review, @lilyminium . 
I agree with your comments. I will merge this as it is and will adapt them in the refactoring PR #34. 
Thanks.

Issue #32, opened 2021-06-24
Refactor `def_all_beads` and add test using toy model.


ojeda-e:
To test mapping (grid) using `pytest.mark.parametrize` as suggested in [PR #30 ](https://github.com/MDAnalysis/membrane-curvature/pull/30#pullrequestreview-687924575), refactor `def_all_beads` to 

- [ ] replace MDtraj by MDAnalysis.
- [ ] Pass arguments as `np.array` instead of topologies. 
Also add:
- [ ] Test function using toy model in small grid of 9 lipids in a 3x3 grid, with x values of 0, 1, 2, and y values of 0, 1, 2.

Issue #28, opened 2021-06-18
Write test for `core_fast` in `mods.py`


ojeda-e:
Write test for `core_fast` function in `mods.py`. This will cover lines 233-241.

This issue will help to close #17

ojeda-e:
Issue resolved with the refactoring introduced in PR #39

PR #34, opened 2021-06-24
Initial refactor and test added [Fixes #33]

### LINKED ISSUE #16: Replace `mdtraj` by `MDAnalysis`
The initial version of the code uses the `mdtraj` library.
Next step is to replace all functions from `mdtraj` by `MDAnalysis` modules and adjust the code wherever is necessary.

Changes required in

- [ ] `core.py`
- [ ] `/lib/mods.py`.

### LINKED ISSUE #33: Refactor `core_fast_leaflet` and `core_fast` and add test using toy model.
To test coordinate mapping and average of `z` values in grid using `pytest.mark.parametrize` as suggested in PR #30, refactor `core_fast_leaflet` and `core_fast` to

- [x] replace MDtraj by MDAnalysis.
- [x] Pass arguments as np.array / tuples instead of topologies.

Also, add:
- [x] Test function using toy model in small grid of 9 lipids in a 3x3 grid, with x values of 0, 1, 2, and y values of 0, 1, 2.
- [x] Test function using toy model in a small grid of 16 lipids in a 4x4 grid, with x values of 0, 1, 2,3 and y values of 0, 1, 2,3. Use different `z` values of beads for this test.

ojeda-e:
Helps to fix #33.
In this first version of the refactored code, the calculation of curvature doesn't evaluate per leaflet, and instead, it runs for a surface defined by selected atoms. (More general)

Changes in this PR:
- [x] Surface derived from direct selection using `MDAnalysis`.
- [x] Deleted functions `def_all_beads`, `curvature` and `core_fast`.
- [x] Function `grid_map` to map coordinates to grid added.  
- [x] In `grid_map` arguments as np.array / tuples instead of topologies. 
- [x] Replaced `MDtraj` by `MDanalysis` 100%.

As suggested [here](https://github.com/MDAnalysis/membrane-curvature/pull/30#pullrequestreview-687924575), tests with toy model using `pytest.mark.parametrize` added:
- [x] `test_grid_map_small_9grid`,  using toy model in small grid of 9 lipids in a 3x3 grid, with x values of 0, 1, 2, and y values of 0, 1, 2.
- [x] `test_grid_map_25grid`, using toy model in small grid of 25 lipids in a 5x5 grid, with x values of 0, 1, 2,3,4 and y values of 0, 1, 2,3,4.

This  PR may also fix
- #27   since function `def_all_beads` was deleted after replacing `MDtraj by `MDAnalysis`. 
- #28   since function `core_fast` was deleted after refactoring. 
- #32  since function `def_all_beads` was deleted after refactoring and replaced by direct selection using MDAnalysis`.

Edit: This PR also fixes #16 .


membrane_curvature/core.py
```diff
@@ -5,60 +5,42 @@
...
-    lipid_types = ['POPC', 'POPE']
-
-    # 4. head indexes
-    head_index = [iil, jjl, jju]
-
-    # 5. Assign topology
-    topology = md.load(grofile).topology
-
-    # 6. Populate universe with coordinates and trajectory
+    # 2. Populate universe with coordinates and trajectory
```
orbeckst, line 25: MDAnalysis is not restricted to GRO format files. I'd just call it *topology* and *trajectory*.

membrane_curvature/core.py
```diff
@@ -5,60 +5,42 @@
...
     u = mda.Universe(grofile, trjfile)
 
-    # 6.1 Set grid: Extract box dimension from MD sim,
+    # 3 Set grid: Extract box dimension from MD sim,
     # set grid max width, set number of unit cells
     box_size = u.dimensions[0]
```
orbeckst, line 30: This gives you Lx only. Do you assume square X-Y ?

orbeckst: Maybe at least
```python
box_size = max(u.dimensions[0], u.dimensions[1])
```

If you assume orthorhombic boxes then consider just failing here if you encounter a triclinic box.

membrane_curvature/lib/mods.py
```diff
@@ -1,150 +1,44 @@
...
-    topology: obj
-        Topology from grofile.
-    Returns
-    -------
-    [ element_1, ..., element_n] : list
-        List with elements used as a reference to derive membrane surface.
+    x: float
+        Value of x coordinate
+    y: float
+        Value of y coordinate
```
orbeckst, line 13: Docs (x,y) do not seem to agree with function signature (coords, factor)

membrane_curvature/lib/mods.py
```diff
@@ -1,150 +1,44 @@
...
-    name : str,
-        name of pickle file.
-    dict_ : dict like {index -> value}
-        dictionary to export.
 
-    Returns
-    -------
-        Returns a pickled dictionary.
+    index_grid_l = int(abs(coords[0]) * factor)
+    index_grid_m = int(abs(coords[1]) * factor)
```
orbeckst, line 22: Is coords a single coordinate (essentially a 3-tuple)?

If so then you should keep in mind for later that you can almost certainly gain quite a bit of performance by doing the grid indexing operation on all coordinates at once (there's code related to `np.histogramdd` that you can probably use... but not in this PR!)

membrane_curvature/lib/mods.py
```diff
@@ -160,27 +54,24 @@ def core_fast_leaflet(z_Ref, leaflet, traj, jump, n_cells, lipid_types,
 
     grid_count = np.zeros([n_cells, n_cells])
 
-    for frame in range(0, traj.n_frames, jump):
```
orbeckst, line 163: What was the `jump` argument good for?

membrane_curvature/lib/mods.py
```diff
@@ -160,27 +54,24 @@ def core_fast_leaflet(z_Ref, leaflet, traj, jump, n_cells, lipid_types,
...
 
         factor = np.float32(n_cells / max_width)
 
-        for lipid_type in lipid_types:
-
-            for bead in lipid_ref[leaflet][lipid_type]:
-
-                x, y, z = traj.xyz[frame, bead, :]
+        for bead in selection:
+            x, y, z = bead.position/10
```
orbeckst, line 65: Why division by 10?

What units is your code working in?

membrane_curvature/lib/mods.py
```diff
@@ -160,27 +54,24 @@ def core_fast_leaflet(z_Ref, leaflet, traj, jump, n_cells, lipid_types,
...
         grid_2 = np.zeros([n_cells, n_cells])
 
         factor = np.float32(n_cells / max_width)
 
-        for lipid_type in lipid_types:
-
-            for bead in lipid_ref[leaflet][lipid_type]:
-
-                x, y, z = traj.xyz[frame, bead, :]
+        for bead in selection:
```
orbeckst, line 64: Once you start optimizing for performance, you most likely will want to completely eliminate this for-loop.

membrane_curvature/lib/mods.py
```diff
@@ -160,27 +54,24 @@ def core_fast_leaflet(z_Ref, leaflet, traj, jump, n_cells, lipid_types,
...
+            x, y, z = bead.position/10
 
-                l = int(abs(x) * factor)
-                m = int(abs(y) * factor)
+            index_l, index_m = grid_map((x, y), factor)
```
orbeckst, line 67: personally, I'd have kept `l` and `m` — they are only used here and I find the more mathematical notation clearer, but different people have different preferences.

membrane_curvature/lib/mods.py
```diff
@@ -160,27 +54,24 @@ def core_fast_leaflet(z_Ref, leaflet, traj, jump, n_cells, lipid_types,
...
 
-                try:
-                    grid_1[l, m] += z
-                    grid_2[l, m] += 1
+            try:
+                grid_1[index_l, index_m] += z
+                grid_2[index_l, index_m] += 1
```
orbeckst, line 71: On the other had, `grid_1` and `grid_2` could do with more expressive names, like `z_grid` and `counts`/`n_grid` ... anything but numbers ;-)

membrane_curvature/lib/mods.py
```diff
@@ -160,27 +54,24 @@ def core_fast_leaflet(z_Ref, leaflet, traj, jump, n_cells, lipid_types,
...
 
-                except BaseException:
-                    pass
+            except ValueError:
+                print("Atom outside grid boundaries. Skipping atom {}".format(bead))
```
orbeckst, line 74: We tend to avoid `print` in production code as it's easy overlooked and people often don't like it when their screen is filled with random messages. If it needs to be printed then we use a logger. If you need to know provide warnings to the user then you can also build up the information in a list and then outside the loop summarize what happened in a message.

For right now, you can leave the `print` but down the line you'll want to think about hoe to communicate with the user.

membrane_curvature/lib/mods.py
```diff
@@ -160,27 +54,24 @@ def core_fast_leaflet(z_Ref, leaflet, traj, jump, n_cells, lipid_types,
...
 
         for i, j in it.product(range(n_cells), range(n_cells)):
             if grid_2[i, j] > 0:
```
orbeckst, line 77: Also amenable to array arithmetic for speedups (later...)

Same for the for loop below.

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -3,239 +3,90 @@
...
+from ..lib.mods import core_fast_leaflet, curvature, mean_curvature, gaussian_curvature, grid_map
 import numpy as np
 from numpy.testing import assert_almost_equal
 import MDAnalysis as mda
-from membrane_curvature.tests.datafiles import GRO_MEMBRANE_PROTEIN, XTC_MEMBRANE_PROTEIN, GRO_PO4, XTC_PO4
+from membrane_curvature.tests.datafiles import (GRO_PO4_SMALL, XTC_PO4_SMALL)
 
 # Reference data from datafile
 MEMBRANE_CURVATURE_DATA = {
-    'mean_curvature': np.array([[0.06396822, 0.2254189, 0.22399141, 0.02498465, 0.05090628,
```
orbeckst, line 22: Why are you removing/changing tests here? This seems dangerous. How do you know that your changes did not introduce bugs? 

For this PR I would have expected _no changes_ to the tests.

orbeckst (commented):
Good progress! I have some general comments.

My main question is why there are so many changes to the tests in this PR. Once you start doing big refactoring like changing from mdtraj to mda, I would have expected the previous tests to be still in place _and passing_. (I might not be fully up-to-date on your changes to the tests but I'd still expect any changes to be tests before this PR.)

membrane_curvature/lib/mods.py
ojeda-e: Interesting, this is another type of detail pep8 shouts at me to tell me that is too vague and I should do it more specific. The previous versions were `l` and `m`.

membrane_curvature/lib/mods.py
ojeda-e: `jump` was to skip frames. I was using `jump = 1` for all the tests so I decided it was not very useful at this point either.

ojeda-e:
Thanks for your comments @orbeckst
There are two main reasons for all the changes here introduced.
- The first one and most relevant, is that in the previous version of my code, topologies and trajectories from`MDtraj` were parsed in three of the functions: `def_all_beads`, `core_fast_leaflets`, and `core_fast`.
With the change from `MDtraj` to `MDAnalysis`, the  arguments of  `def_all_beads` disappeared, `core_fast` `core_fast_leaflets` significantly changed. 
- The second reason is that in previous PRs (see PR #31 and #30) the tests I was asked to provide were with dummy coordinates, to be more readable. Which is completely ok, but makes more sense only after refactoring. 
As I mentioned above, In the previous version of the code, the arguments parsed in 3 of the functions were `MDtraj` topologies and trajectories that I can't build with numpy arrays. Only after this refactoring, and having MDAnalysis instead, that type of test makes more sense.
- Additionally, since the previous PRs I was asked to have tests using dummy coordinates and the ones I provided with grofiles were getting too complicated to read, I simplify the tests in this new version of the code.

In the spirit of moving forward, I added changes in this PR that wrap up the previous limitations and allows me to progress. If further tests are needed, now with this refactored version providing dummy coordinates will be possible. I feel I haven't made progress because of the tests that have been asked previously, which were going to change no matter what with this PR.
If it's better, I can put the big systems tests for two of the functions, `mean_curvature` and `gaussian_curvature`, which are essentially the only two functions that remain the same after refactoring.

membrane_curvature/core.py
ojeda-e: Yes, this is one of the limitations and one of the next issues to solve. Issue #35 added.

membrane_curvature/lib/mods.py
ojeda-e: Yes, you are right and thanks for this catch. I'll update it.

membrane_curvature/lib/mods.py
ojeda-e: In `nm` :(

membrane_curvature/tests/test_mdakit_membcurv.py
ojeda-e: I will put them back, I just thought of simplifying the tests by providing smaller arrays. But I see your point.

ojeda-e: Previous tests`test_gaussian_curvature` and `test_mean_curvature`(now called `test_gaussian_curvature_all`and `test_mean_curvature_all`, repectively) added back and passing.

membrane_curvature/lib/mods.py
```diff
@@ -158,38 +52,35 @@ def core_fast_leaflet(z_Ref, leaflet, traj, jump, n_cells, lipid_types,
...
-                grid_count[i, j] += 1
+            if grid_norm_unit[i, j] > 0:
+                z_Ref[i, j] += grid_z_coordinates[i, j] / grid_norm_unit[i, j]
+                grid_count_frames[i, j] += 1
 
     for i, j in it.product(range(n_cells), range(n_cells)):
-        if grid_count[i, j] > 0:
-            z_Ref[i, j] /= grid_count[i, j]
+        if grid_count_frames[i, j] > 0:
+            z_Ref[i, j] /= grid_count_frames[i, j]
```
lilyminium, line 83: I see that you're first creating an average surface before computing the curvature. On a scientific level, maybe we should consider treating each surface for each frame separately? Otherwise, if you have a membrane that undulates such that a patch has ~50% positive curvature and 50% negative curvature over the frames, the Gaussian curvature is ultimately computed on a flat surface?

lilyminium (changes requested):
This is a great start to refactoring, @ojeda-e -- the new grid_map tests are much more digestible!

I have a few notes on the refactor. I think it's worth spending some time thinking about how you want to break your MDTraj code and put it back together in an MDAnalysis format. You've swapped MDTraj trajectories for MDAnalysis universes in `core_fast_leaflet`, which is a good start, but IMO an iterative kind of refactor will constrain potential designs because of trying to be similar to the previous iteration.

Instead, this seems to be a good time to reduce complexity and start separating your functions into more modular ones, where each function does only one main thing. Then it will be easier to put them together in convenient ways as we try to work out the best design for the MembraneCurvature AnalysisBase class. (and easier to test!)

Currently `core_fast_leaflet` does three things:

* gets positions for each atom in the atom group for each frame
* identifies the grid cell for each coordinate (currently grid_map)
* calculates the average z for the atom group

Could you please break this down into three functions, and write a new function that simply calls them (just for the regression tests?) I would be keen to have these as divorced from either MDTraj or MDAnalysis as possible and work only with numpy arrays, which are generally more versatile. The overall function could grab the positions from the AtomGroup and feed it into the functions instead.

In addition, some unofficial principles of design from personal experience:

**Side-effecting functions**

Currently `core_fast_leaflet` is a [side-effecting function](https://en.wikipedia.org/wiki/Side_effect_(computer_science)). This means that it modifies a variable, `z_Ref`, outside its local environment. In functional programming, where you write functions (as opposed to class/objects in object-oriented programming), it's generally good practice to *avoid* side-effecting functions. This [StackOverflow post](https://softwareengineering.stackexchange.com/questions/15269/why-are-side-effects-considered-evil-in-functional-programming) summarises it better than I could, but in general side-effect free functions are easier to test, easier to parallelize, and easier to cobble together into larger functions.

To write a side-effect free function, `core_fast_leaflet` would be given input arguments that it doesn't modify, and return an output value. So the `test_core_fast_lealets` function could look like:

```diff
def test_core_fast_leaflets():
    n_cells, max_width = 3, 3
---    z_calc = np.zeros([n_cells, n_cells])
    u = mda.Universe(GRO_PO4_SMALL, XTC_PO4_SMALL)
    selection = u.select_atoms('index 0:3')
---    core_fast_leaflet(u, z_calc, n_cells, selection, max_width)
+++    z_calc = core_fast_leaflet(u, n_cells, selection, max_width)
    for z, z_test in zip(MEMBRANE_CURVATURE_DATA['grid']['small']['upper'], z_calc):
        print(z, z_test)
        assert_almost_equal(z, z_test)
```


**The principle of least astonishment**

The [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) basically advises that you use your code to create user expectations, and then try your best to fulfil them. This comes in many forms, such as naming your variables and functions clearly (more on that below). It also means trying to follow the conventions of other packages as best you can. This is a bit nitpicky, but one example is the input arguments to `np.zeros`. The documentation actually specifies that the shape should be an integer or a tuple of integers. In most contexts, code that looks like `np.func([x, y])` means that `np.func` is operating on a list of values that will be converted to an array. In contrast, shape is commonly specified as a tuple `(x, y)` so it's much less likely to be misunderstood.

The type of input arguments to `np.zeros` is of course a very small issue. I just bring it up now because as I said above, refactoring is a good time to start thinking about what kind of inputs and outputs you want your functions to take, as that is a key part of API design.

Even more nitpicky: parameter ordering is also part of designing a function. There isn't really much formal advice on how to order parameters in a function (although I found [this Stackoverflow post](https://softwareengineering.stackexchange.com/questions/101346/what-is-best-practice-on-ordering-parameters-in-a-function)). This is not important right now, but for example with `core_fast_leaflet`, as a user, I would be surprised at the ordering of `universe, z_Ref, n_cells, selection, max_width`. That's because I think of universes and AtomGroups as very similar things (in fact, you don't necessarily need universe as an argument, as you could get the `universe = selection.universe`). In addition, `n_cells` and `max_width` both relate to the same thing: `factor`. It feels more intuitive to me to call `core_fast_leaflet(universe, selection, z_Ref, n_cells, max_width)` and have those arguments grouped together.

**Naming functions**

Part of PLOA is naming things clearly, so future developers (or you in 2 months!) can read your code with minimal effort. PEP8 has a [whole section on naming conventions](https://www.python.org/dev/peps/pep-0008/#naming-conventions). In general, it's typically a good idea to have a verb in your function name. For example, I am not sure what `grid_map` does just by looking at its name. If it were more explicit and map were a verb, e.g. `map_coordinates_to_grid`, that's much more immediately obvious. It's up to you how verbose you would like to make it; other candidates could be `map_coordinates_to_array`, `map_coordinates_to_grid_array`, `get_coordinate_cell`.

The kind of verb can hint at the expected output, such as `get_coordinate_cell` above. For example, I would expect functions that start with `is_` or `has_` (e.g. `has_atoms`, `is_negative`) to return either True or False.

ojeda-e:
Hi @lilyminium Lily, thanks for your review. Regarding this comment (and below)

> Currently `core_fast_leaflet` does three things:
> 
> * gets positions for each atom in the atom group for each frame
> * identifies the grid cell for each coordinate (currently grid_map)
> * calculates the average z for the atom group
> 
> Could you please break this down into three functions, and write a new function that simply calls them (just for the regression tests?) I would be keen to have these as divorced from either MDTraj or MDAnalysis as possible and work only with numpy arrays, which are generally more versatile. The overall function could grab the positions from the AtomGroup and feed it into the functions instead.

If it works for you,  I'll add these remarks as new issues instead of adding them in this PR. The changes are getting a bit too long. Would that work?
If yes, then it'll be happy to submit a PR with the requested changes.

PR #38, opened 2021-06-27
Requested changes to initial refactor

### LINKED ISSUE #33: Refactor `core_fast_leaflet` and `core_fast` and add test using toy model.

ojeda-e:
This PR fixes #33. 

Changes included:

- [x] `pytest.mark.xfail` test of negative coordinates in `grid_map`. 
- [x]  9- and 25-grid combined into one function with `pytest.parametrize`.
- [x] `Test_mapper` changed by adding assertion of dummy coordinates.
- [x] Test with `grid_map` using 6 unit cells in grid of 9 lipids as shown below:

```
^   o ___ o ___ o ___ o ___ o ___ o ___ |
|   | (6) |     | (7) |     | (8) |     |
|   o ___ o ___ o ___ o ___ o ___ o ___ |
|   |     |     |     |     |     |     |
|   o ___ o ___ o ___ o ___ o ___ o ___ |
y   | (3) |     | (4) |     | (5) |     | 
|   o ___ o ___ o ___ o ___ o ___ o ___ |
|   |     |     |     |     |     |     |
|   o ___ o ___ o ___ o ___ o ___ o ___ |
|   | (0) |     | (1) |     | (2) |
v   o ___ o ___ o ___ o ___ o ___ o ___ |
    0.   0.5    1.   1.5.   2.   2.5.   3.
    <----------------- x --------------->
```
For grid with 9 lipid types:
Bead  0  mapped to [ 0 0 ], with Coordinates ( 0 , 0 )
Bead  1  mapped to [ 2 0 ], with Coordinates ( 1 , 0 )
Bead  2  mapped to [ 4 0 ], with Coordinates ( 2 , 0 )
Bead  3  mapped to [ 0 2 ], with Coordinates ( 0 , 1 )
Bead  4  mapped to [ 2 2 ], with Coordinates ( 1 , 1 )
Bead  5  mapped to [ 2 2 ], with Coordinates ( 1 , 1 )
Bead  6  mapped to [ 0 4 ], with Coordinates ( 0 , 2 )
Bead  7  mapped to [ 2 4 ], with Coordinates ( 1 , 2 )
Bead  8  mapped to [ 4 4 ], with Coordinates ( 2 , 2 )

and the equivalent to the grid of 25 lipids.

This PR may also fix:
#27  since function `def_all_beads` was deleted after replacing `MDtraj` by `MDAnalysis`.
#28 since function `core_fast` was deleted after refactoring.
#32 since function `def_all_beads` was deleted after refactoring and replaced by direct selection using `MDAnalysis`.
and #16 

Questions:
(possible minor change)
- In line 172 and 177, I could simplify to `factor = 1`
and then in lines 174 and 179, respectively.
`assert grid_map(dummy_coord, 1) == dummy_coord`
`assert grid_map(dummy_coord, 2) == (dummy_coord[0]*2, dummy_coord[1]*2)`


ojeda-e:
Hi @lilyminium, could I have a review in this PR, please :) 
Thanks.

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -155,33 +155,43 @@ def test_core_fast_leaflets():
...
+    # dummy coordinates (x,y) in grid of 5x5
     ((0, 0), (1, 0), (2, 0), (3, 0), (4, 0),
      (0, 1), (1, 1), (2, 1), (3, 1), (4, 1),
      (0, 2), (1, 2), (2, 2), (3, 2), (4, 2),
      (0, 3), (1, 3), (2, 3), (3, 3), (4, 3),
-     (0, 4), (1, 4), (2, 4), (3, 4), (4, 4)),
+     (0, 4), (1, 4), (2, 4), (3, 4), (4, 4)))])
+def test_grid_map_grids(dummy_systems):
+    # map to grid of 3 unit cells
+    for grid in dummy_systems:
```
fiona-naughton, line 171: `pytest.mark.parametrize` should take care of running the test in turn for each dummy system in the list you provide, so you shouldn't need this line (or 176). I believe you'll need to remove the outer set of `()` on the dummy system list to get that working properly (so it passes in the individual systems rather than the full list).

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -155,33 +155,43 @@ def test_core_fast_leaflets():
...
+        factor = np.float32(np.sqrt(len(grid)) / np.sqrt(len(grid)))
```
fiona-naughton, line 172: As you mentioned, it makes sense to simplify and just pass `factor=1`/`factor=2` directly into `grid_map` rather than this

membrane_curvature/tests/test_mdakit_membcurv.py
ojeda-e: Thanks, I missed this. Fixed :)

ojeda-e:
Thanks for the comments @fiona-naughton. Changes added.

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -155,33 +155,41 @@ def test_core_fast_leaflets():
...
-])
-def test_grid_map_small_9grid(dummy_coordinates, test_mapper, n_cells, max_width):
-    factor = np.float32(n_cells / max_width)
-    for dummy_coord in dummy_coordinates:
-        assert test_mapper(dummy_coord) == grid_map(dummy_coord, factor)
-
-
-@pytest.mark.parametrize('dummy_coordinates, test_mapper, n_cells, max_width', [(
-    # dummy coordinates (x,y)
+@pytest.mark.parametrize('grid', [(
```
lilyminium, line 158:
```suggestion
@pytest.mark.parametrize('factor', [1, 2])
@pytest.mark.parametrize('grid', [(
```

One fun thing about `parametrize` is that you can next them. For example, adding this `factor` line will result in four cases:

* factor = 1, grid = first grid
* factor = 1, grid = second grid
* factor = 2, grid = first grid
* factor = 2, grid = second grid

This can let you reduce the test itself a little, because now each iteration only tests 1 factor value and 1 grid value.

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -155,33 +155,41 @@ def test_core_fast_leaflets():
...
+    (0, 2), (1, 2), (2, 2)),
+    # dummy coordinates (x,y) in grid of 5x5
     ((0, 0), (1, 0), (2, 0), (3, 0), (4, 0),
      (0, 1), (1, 1), (2, 1), (3, 1), (4, 1),
      (0, 2), (1, 2), (2, 2), (3, 2), (4, 2),
      (0, 3), (1, 3), (2, 3), (3, 3), (4, 3),
-     (0, 4), (1, 4), (2, 4), (3, 4), (4, 4)),
+     (0, 4), (1, 4), (2, 4), (3, 4), (4, 4))])
+def test_grid_map_grids(grid):
+    # map to grid of 3 unit cells
```
lilyminium, line 170: These comments are now not accurate :)

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -155,33 +155,41 @@ def test_core_fast_leaflets():
...
+    factor_1 = 1
+    for dummy_coord in grid:
+        assert grid_map(dummy_coord, factor_1) == dummy_coord
```
lilyminium, line 173: Looking at this for loop, I am a little confused as to why you are passing in a grid at all. Why not just have this test go:

```python

@pytest.mark.parametrize('factor', [1, 2])
@pytest.mark.parametrize('dummy_coord', [
	(0, 0),
	(0, 1),
	... # and so on
])
def test_grid_map(factor, dummy_coord):
	expected_cell = (dummy_coord[0] * factor, dummy_coord[1] * factor)
	assert grid_map(dummy_coord, factor) == expected_cell
```

The fact that the coordinates are from a grid is not actually relevant to this test.

The same applies to the test below.

lilyminium (changes requested):

lilyminium (changes requested):
Thanks for making this change! Could you please just remove repeated values like (0, 0) -- although to be honest the second 5x5 grid isn't necessary.

ojeda-e:
> Thanks for making this change! Could you please just remove repeated values like (0, 0) -- although to be honest the second 5x5 grid isn't necessary.

I added the 5x5 as suggested in previous PRs. If you are ok with it with different values excluding (0,0). @lilyminium  Maybe inverted values? As in 
```
(4, 0), (3, 0), (2, 0), (1, 0), (1, 1),
(4, 1), (3, 1), (2, 1), (1, 1), (0, 2),
(4, 2), (3, 2), (2, 2), (1, 2), (0, 3),
(4, 3), (3, 3), (2, 3), (1, 3), (0, 5),
(4, 4), (3, 4), (2, 4), (1, 4), (0, 4)
```
I would prefer to keep it even if not necessary.

lilyminium:
> Maybe inverted values?

I think those are just the same values in a different order, but sure, we can leave the grid in as-is so it's easier to modify for future tests :)

lilyminium (approved):

Issue #36, opened 2021-06-25
Add pbc conditions to in mapping coordinates to grid.


ojeda-e:
Periodic Boundary Conditions (PBC) are not applied when mapping coordinates to grid (`grid_map`).

To fix this issue:
- [ ] Add PBC conditions to map coordinates.
- [ ] Add test using dummy coordinates to map values that exceed by defect or excess the boundaries of the grid.

For example: 
- System with 9 beads with  coordinates of `x` -1, 0, 1 and `y` -1, 0 ,1
- System with 16 beads with  coordinates of `x` -2, -1, 0, 1, and `y` -2, -1, 0 ,1, 2.

richardjgowers:
Try and avoid implementing this yourself, check out lib.distances.apply_PBC

ojeda-e:
I was considering to apply PBC is in `AnalysisBase`, in `__init__`:

```
        # Apply PBC conditions
        if self.pbc == True:
            self.ag.wrap()
        else:
            warnings.warn(" `PBC == False` may result in inaccurate calculation "
                          "of membrane curvature. Surfaces will be derived from "
                          "a reduced number of atoms.")
```

I think this is a better approach than applying it in `_single_frame` or in the `get_z_surface` function (but I may be wrong). If you are ok with this approach, I can add it with respective tests to the current PR #41  and this issue may get fixed. 

What do you think @IAlibay, @lilyminium, @orbeckst, @fiona-naughton?

IAlibay:
> I was considering to apply PBC is in `AnalysisBase`, in `__init__`:
> 
> ```
>         # Apply PBC conditions
>         if self.pbc == True:
>             self.ag.wrap()
>         else:
>             warnings.warn(" `PBC == False` may result in inaccurate calculation "
>                           "of membrane curvature. Surfaces will be derived from "
>                           "a reduced number of atoms.")
> ```
> 
> I think this is a better approach than applying it in `_single_frame` or in the `get_z_surface` function (but I may be wrong). If you are ok with this approach, I can add it with respective tests to the current PR #41 and this issue may get fixed.
> 
> What do you think @IAlibay, @lilyminium, @orbeckst, @fiona-naughton?

I might be misunderstanding what you're trying to do here, but isn't `ag.wrap()` only applied to that timestep as it is being accessed? I.e. when you traverse through the trajectory the `wrap()` stops being applied right?

orbeckst:
[...]

Issue #35, opened 2021-06-25
Define grid size for different box shapes.


ojeda-e:
As pointed out by @orbeckst, after initial refactor in #34 the grid is assumed to be square. The grid should accept at least rectangular shapes.

To do:
- [ ] Define the size of the grid by at least `x` and `y`. 
Suggestion: `box_size = max(u.dimensions[0], u.dimensions[1])`
- [ ] Add `xfail` for  orhtorombic/triclinic box.

ojeda-e:
I think we I can close this issue since we have 
`self.x_range = x_range if x_range else (0, universe.dimensions[0])`
`self.y_range = y_range if y_range else (0, universe.dimensions[1])` 

in `base.py` and respective tests with rectangular box sizes.

Do you agree @orbeckst @IAlibay @lilyminium ? Am I missing something here? Thanks.

lilyminium:
If you add a rectangular grid test then yeah I think this is covered!

Issue #22, opened 2021-06-16
Add test for the generation of z_ref (core_fast_leaflet) on toy system


lilyminium:
The code to transform xy coordinates to grid cells is important and modular enough that you could keep it as a function in the AnalysisBase refactor. Therefore, it should get tested extensively to make sure that it is correct, especially as the regression tests for the mean and gaussian curvature in #13 depend on z_ref coordinates that were presumably generated this way. Writing the test could be done before, or in parallel, with starting to outline the AnalysisBase re-factor.

Instead of testing the generation of z_ref on the GRO file you have uploaded, a simpler test that is more easy to humanly verify would be a small toy system of fake coordinates (maybe 5-12 coordinates per axis for a suitable corresponding number of cells). You could use `pytest.mark.parametrize` to test multiple inputs (or you could have separate tests for each one if that's more your style!). You might need to have at least two tests though, one for tests that currently pass and one for tests that are expected to fail (marked with XFAIL, e.g. those with PBC) These inputs could include:

* negative coordinates
* coordinates right on the bounds of each cell
* a square input (e.g 8 x 8)
* a rectangular input (e.g. 8 x 12)
* coordinates higher than the box size (e.g. for unwrapped trajectories)
* the scenario of z-coordinates on the bottom of the box and requiring PBC

For example, I would check that a system of [(-1, 0), (0, 0), (1, 0), (2, 0), (3, 0)] gives a grid with the (-1, 0) coordinate last and only one y-row. This smaller system is much easier to understand than a real membrane, and defines expectations such as PBC.

ojeda-e:
I think this issue is fixed with the tests added in PR #48  
Would you please confirm this is ok, @lilyminium?

lilyminium:
If you add a test for a rectangular grid then I think it's all covered!

ojeda-e:
Thanks. I added one rectangular test as part of a pytest parametrize to check MembranCurvature.results.average_z_surface, block in lines 307-324.
If that’s not enough I can add a couple for mc.results.mean and mc.results.gaussian. :)

Sent from my iPhone
________________________________

PR #40, opened 2021-06-29
Added `derive_surface` and `get_positions` with tests.

ojeda-e:
Function `core_fast_leaflets` split into three functions:

- [x] get positions for each atom in the atom group for each frame.
- [x]  identify the grid cell for each coordinate.
- [x] calculate the average z for the atom group.

Tests added:
- [x] `test_get_positions` using dummy coordinates for beads 0 to 8, all of them with `z=10`:
```
o ______ o _____ o _______ |
|   (6)  |  (7)   |   (8)  |
o ______ o _____ o _______ |
|   (3)  |  (4)   |   (5)  |
o _______o ______ o ______ |
|   (0)  |  (1)   |   (2)  |
o ______ o ______ o ______ |
```
Using the same number of beads in grids, 

- [x] `test_avg_unit_cell` added for two systems.
1.  z values of `z=10`
2. z values as below:
```
o ______ o _____ o _______ |
| (z=10) | (z=20) | (z=30) |
o ______ o _____ o _______ |
| (z=10) | (z=20) | (z=30) |
o _______o ______ o ______ |
| (z=10) | (z=20) | (z=30) |
o ______ o ______ o ______ |
```
3. z values of `z=10` and number of beads per unit cell as shown below:
```
o ____ o ____ o ___ |
|   2  |  1   |  1  |
o ____ o ____ o ___ |
|   1  |  2   |  1  |
o ____ o ____ o ___ |
|   1  |  1   |  2  |
o ____ o ____ o ___ |
```

- [x] `test_derive_surface` added for same `dummy_coordinates` as in `test_get_positions`.

### DIFF
--- a/membrane_curvature/lib/mods.py
+++ b/membrane_curvature/lib/mods.py
@@ -1,27 +1,33 @@
 import numpy as np
 
 
-def derive_surface(n_cells, selection, max_width_x, max_width_y):
+def derive_surface(n_cells_x, n_cells_y, selection, max_width_x, max_width_y):
     """
     Derive surface from AtomGroup positions.
 
     Parameters
     ----------
-    n_cells : int.
-        number of cells in the grid of size `max_width`.
+    n_cells_x : int.
+        number of cells in the grid of size `max_width_x`, `x` axis.
+    n_cells_y : int.
+        number of cells in the grid of size `max_width_y`, `y` axis.
     selection : AtomGroup.
         AtomGroup of reference selection to define the surface
         of the membrane.
-    max_width : float.
-        Maximum width of simulation box. (Determined by simulation box dimensions)
+    max_width_x: float.
+        Maximum width of simulation box in x axis. (Determined by simulation box dimensions)
+    max_width_y: float.
+        Maximum width of simulation box in y axis. (Determined by simulation box dimensions)
 
     Returns
     -------
-    Returns set of z coordinates 
+    z_coordinates: numpy.ndarray
+        Average z-coordinate values. Return Numpy array of floats of
+        shape `(n_cells_x, n_cells_y)`.
 
     """
     coordinates = selection.positions
-    return get_z_surface(coordinates, n_x_bins=n_cells, n_y_bins=n_cells,
+    return get_z_surface(coordinates, n_x_bins=n_cells_x, n_y_bins=n_cells_y,
                          x_range=(0, max_width_x), y_range=(0, max_width_y))
 
 
@@ -31,21 +37,23 @@ def get_z_surface(coordinates, n_x_bins=10, n_y_bins=10, x_range=(0, 100), y_ran
 
     Parameters
     ----------
-    coordinates : tuple.
-        Numpy.ndarray with shape=(n_atoms, 3).
+    coordinates : numpy.ndarray 
+        Coordinates of AtomGroup. Numpy array of shape=(n_atoms, 3).
     n_x_bins : int.
-        Number of bins in grid in the x dimension. 
+        Number of bins in grid in the `x` dimension. 
     n_y_bins : int.
-        Number of bins in grid in the y dimension. 
-    x_range : tuple
-        Range of indexes in grid in the x dimension with shape=(0, max_width_x).
-    y_range : tuple
-        Range of indexes in grid in the y dimension with shape=(0, max_width_y)
+        Number of bins in grid in the `y` dimension. 
+    x_range : tuple of (float, float)
+        Range of indexes in grid in the `x` dimension with shape=(2,).
+    y_range : tuple of (float, float)
+        Range of indexes in grid in the `y` dimension with shape=(2,).
 
 
     Returns
     -------
-    Returns surface derived from z coordinates in grid of `n_x_bins` x `n_y_bins` dimensions.
+    z_surface: numpy.ndarray 
+        Surface derived from set of coordinates in grid of `x_range, y_range` dimensions.
+        Returns Numpy array of floats of shape (`n_x_bins`, `n_y_bins`)
 
     """
 
@@ -95,9 +103,8 @@ def avg_unit_cell(z_ref, grid_z_coordinates, grid_norm_unit):
 
     """
 
-    normed = grid_norm_unit > 0
+    grid_norm_unit = np.where(grid_norm_unit > 0, grid_norm_unit, np.nan)
     z_ref = grid_z_coordinates / grid_norm_unit
-    z_ref[~normed] = np.nan
 
     return z_ref
 
--- a/membrane_curvature/lib/mods.py
+++ b/membrane_curvature/lib/mods.py
@@ -44,10 +44,9 @@ def get_z_surface(coordinates, n_x_bins=10, n_y_bins=10, x_range=(0, 100), y_ran
     n_y_bins : int.
         Number of bins in grid in the `y` dimension. 
     x_range : tuple of (float, float)
-        Range of indexes in grid in the `x` dimension with shape=(2,).
+        Range of coordinates (min, max) in the `x` dimension with shape=(2,).
     y_range : tuple of (float, float)
-        Range of indexes in grid in the `y` dimension with shape=(2,).
-
+        Range of coordinates (min, max) in the `y` dimension with shape=(2,). 
 
     Returns
     -------
--- a/membrane_curvature/lib/mods.py
+++ b/membrane_curvature/lib/mods.py
@@ -13,20 +13,20 @@ def derive_surface(n_cells, selection, max_width):
     selection : AtomGroup
         AtomGroup of reference selection to define the surface
         of the membrane.
-    max_width : int.
-        Maximum width of simulation box.
+    max_width : float.
+        Maximum width of simulation box. (Determined by simulation box dimensions)
 
     Returns
     -------
     Returns set of z coordinates in grid.
 
     """
-    NM_TO_ANGSTROM = 10  # Factor needed temporarily until we can make this code unit-aware or unitless
+
     z_ref = np.zeros((n_cells, n_cells))
     grid_z_coordinates = np.zeros((n_cells, n_cells))
     grid_norm_unit = np.zeros([n_cells, n_cells])
 
-    # max_width *= NM_TO_ANGSTROM # scaling the max width of the box. Will remove
+    
     factor = np.float32(n_cells / max_width)
 
     cell_xy_floor = np.int32(selection.positions[:, :2] * factor)
--- a/membrane_curvature/lib/mods.py
+++ b/membrane_curvature/lib/mods.py
@@ -1,41 +1,5 @@
-import itertools as it
+from ast import NameConstant
 import numpy as np
-
-
-def get_positions(index):
-    """
-    Get positions of bead by index
-
-    Parameters
-    ----------
-    index: Indexed assigned to bead in AtomGroup
-
-    """
-    x, y, z = index
-    return x, y, z
-
-
-def grid_map(coords, factor):
-    """
-    Maps (x,y) coordinates to unit cell in grid.
-
-    Parameters
-    ----------
-    (x, y):  tuple
-        Value of (x, y) coordinates
-    factor:  float
-        Mapping factor to assign grid.
-
-    Returns
-    -------
-    Returns l, m  with l,m as int.
-
-    """
-
-    index_grid_l = int(abs(coords[0]) * factor)
-    index_grid_m = int(abs(coords[1]) * factor)
-
-    return index_grid_l, index_grid_m
 
 
 def derive_surface(n_cells, selection, max_width):
@@ -47,7 +11,7 @@ def derive_surface(n_cells, selection, max_width):
     n_cells : int.
         number of cells in the grid of size `max_width`.
     selection : AtomGroup
-        AtomGroup of reference selected to define the surface
+        AtomGroup of reference selection to define the surface
         of the membrane.
     max_width : int.
         Maximum width of simulation box.
@@ -57,37 +21,49 @@ def derive_surface(n_cells, selection, max_width):
     Returns set of z coordinates in grid.
 
     """
-    z_ref = np.zeros([n_cells, n_cells])
-    grid_z_coordinates = np.zeros([n_cells, n_cells])
+    NM_TO_ANGSTROM = 10  # Factor needed temporarily until we can make this code unit-aware or unitless
+    z_ref = np.zeros((n_cells, n_cells))
+    grid_z_coordinates = np.zeros((n_cells, n_cells))
     grid_norm_unit = np.zeros([n_cells, n_cells])
 
-    factor = np.float32(n_cells / max_width)
+    factor = np.float32(n_cells / max_width * NM_TO_ANGSTROM)
 
-    for bead in selection:
+    cell_xy_floor = np.int32(selection.positions[:, :2] / factor)
+    z_coordinate = selection.positions[:, 2] / factor * NM_TO_ANGSTROM
 
-        x, y, z = get_positions(bead)
-
-        l, m = grid_map((x, y), factor)
+    for (l, m), z in zip(cell_xy_floor, z_coordinate):
 
         try:
             grid_z_coordinates[l, m] += z
             grid_norm_unit[l, m] += 1
 
-        except ValueError:
-            print("Atom outside grid boundaries. Skipping atom {}".format(bead))
+        except IndexError:
+            print("Atom outside grid boundaries. Skipping atom.")
 
-    surface = avg_unit_cell(z_ref, n_cells, grid_z_coordinates, grid_norm_unit)
+    surface = avg_unit_cell(z_ref, grid_z_coordinates, grid_norm_unit)
 
     return surface
 
 
-def avg_unit_cell(z_ref, n_cells, grid_z_coordinates, grid_norm_unit):
+def avg_unit_cell(z_ref, grid_z_coordinates, grid_norm_unit):
+    """
+    Calculates average z coordinate in unit cell
 
-    for i, j in it.product(range(n_cells), range(n_cells)):
-        if grid_norm_unit[i, j] > 0:
-            z_ref[i, j] += grid_z_coordinates[i, j] / grid_norm_unit[i, j]
-        else:
-            z_ref[i, j] = np.nan
+    Parameters
+    ----------
+
+    z_ref: np.array
+        Empty array of (l,m) 
+    grid_z_coordinates: np.array
+        Array of size (l,m) with z coordinates stored in unit cell.
+    grid_norm_unit: np.array
+        Array of size (l,m) with number of atoms in unit cell.
+
+    """
+
+    normed = grid_norm_unit > 0
+    z_ref = grid_z_coordinates / grid_norm_unit
+    z_ref[~normed] = np.nan
 
     return z_ref
 
--- a/membrane_curvature/lib/mods.py
+++ b/membrane_curvature/lib/mods.py
@@ -57,7 +57,6 @@ def get_z_surface(coordinates, n_x_bins=10, n_y_bins=10, x_range=(0, 100), y_ran
 
     """
 
-    z_ref = np.zeros((n_x_bins, n_y_bins))
     grid_z_coordinates = np.zeros((n_x_bins, n_y_bins))
     grid_norm_unit = np.zeros((n_x_bins, n_y_bins))
 
@@ -78,12 +77,12 @@ def get_z_surface(coordinates, n_x_bins=10, n_y_bins=10, x_range=(0, 100), y_ran
         except IndexError:
             print("Atom outside grid boundaries. Skipping atom.")
 
-    z_surface = avg_unit_cell(z_ref, grid_z_coordinates, grid_norm_unit)
+    z_surface = normalized_grid(grid_z_coordinates, grid_norm_unit)
 
     return z_surface
 
 
-def avg_unit_cell(z_ref, grid_z_coordinates, grid_norm_unit):
+def normalized_grid(grid_z_coordinates, grid_norm_unit):
     """
     Calculates average z coordinate in unit cell
 
@@ -104,9 +103,9 @@ def avg_unit_cell(z_ref, grid_z_coordinates, grid_norm_unit):
     """
 
     grid_norm_unit = np.where(grid_norm_unit > 0, grid_norm_unit, np.nan)
-    z_ref = grid_z_coordinates / grid_norm_unit
+    z_normalized = grid_z_coordinates / grid_norm_unit
 
-    return z_ref
+    return z_normalized
 
 
 def gaussian_curvature(Z):
--- a/membrane_curvature/lib/mods.py
+++ b/membrane_curvature/lib/mods.py
@@ -3,8 +3,8 @@ import numpy as np
 
 
 def get_positions(index):
-    """ 
-    Get positions of bead by index 
+    """
+    Get positions of bead by index
 
     Parameters
     ----------
@@ -16,7 +16,7 @@ def get_positions(index):
 
 
 def grid_map(coords, factor):
-    """ 
+    """
     Maps (x,y) coordinates to unit cell in grid.
 
     Parameters
--- a/membrane_curvature/lib/mods.py
+++ b/membrane_curvature/lib/mods.py
@@ -1,16 +1,15 @@
-from ast import NameConstant
 import numpy as np
 
 
-def derive_surface(n_cells, selection, max_width):
+def derive_surface(n_cells, selection, max_width_x, max_width_y):
     """
-    Derive surface from distribution of z coordinates in grid.
+    Derive surface from AtomGroup positions.
 
     Parameters
     ----------
     n_cells : int.
         number of cells in the grid of size `max_width`.
-    selection : AtomGroup
+    selection : AtomGroup.
         AtomGroup of reference selection to define the surface
         of the membrane.
     max_width : float.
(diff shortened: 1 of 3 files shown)


membrane_curvature/lib/mods.py
```diff
@@ -24,66 +38,58 @@ def grid_map(coords, factor):
     return index_grid_l, index_grid_m
 
 
-def core_fast_leaflet(universe, z_Ref, n_cells, selection, max_width):
+def derive_surface(n_cells, selection, max_width):
```
p-j-smith, line 41: Hi @ojeda-e I'm looking forward to being able to use your membrane curvature tool! I'm not sure if this fits your purpose exactly, but you could use [`scipy.stats.binned_statistic_2d`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.binned_statistic_2d.html) to generate a surface of all your beads in one go.

membrane_curvature/lib/mods.py
```diff
@@ -24,66 +38,58 @@ def grid_map(coords, factor):
...
 
-        for bead in selection:
-            x, y, z = bead.position/10
+        try:
+            grid_z_coordinates[l, m] += z
+            grid_norm_unit[l, m] += 1
 
-            l, m = grid_map((x, y), factor)
+        except ValueError:
+            print("Atom outside grid boundaries. Skipping atom {}".format(bead))
```
p-j-smith, line 77: could you do `selection.wrap` to ensure all atoms are within the grid boundaries?

lilyminium (changes requested):
Hi @ojeda-e, I think I had something different in mind when I requested that you only use numpy etc and break it down to basics. Could you have a go at writing the function with the input arguments I suggested? I haven't had a look at tests because I assume they'll change with the new function.

@p-j-smith had a good suggestion with the scipy function. When I said "numpy only" that was probably too restrictive, sorry; totally on me. For future reference, you should consider any package safe if it's something that MDAnalysis considers a ["core dependency"](https://userguide.mdanalysis.org/stable/module_imports.html#list-of-core-module-dependencies). I actually find it preferable to use numpy/scipy functions because they're usually written by people much smarter than me, who've thought about the problem longer than me -- this usually results in faster, better code.

If you try something like the code below, you might make this a very short function! 

```python
x, y, z = coordinates.T
scipy.stats.binned_statistic_2d(x, y, z, statistic="mean", bins=..., range=...)
```

membrane_curvature/lib/mods.py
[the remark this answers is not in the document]
lilyminium: Looking back over previous comments I realise that this directly contradicts https://github.com/MDAnalysis/membrane-curvature/pull/34#pullrequestreview-693038505 . Sorry if I am being confusing and backtracking on that. The motivation for the original comment came from how long the original function was. If the same thing can be accomplished with one line, it doesn't need to be separate functions anymore.

> I would be keen to have these as divorced from either MDTraj or MDAnalysis as possible and work only with numpy arrays, which are generally more versatile. The overall function could grab the positions from the AtomGroup and feed it into the functions instead.

The above point is the more important one that having separate functions -- working with just numpy arrays for the base function.

membrane_curvature/lib/mods.py
```diff
@@ -24,66 +38,58 @@ def grid_map(coords, factor):
...
     """
+    z_ref = np.zeros([n_cells, n_cells])
+    grid_z_coordinates = np.zeros([n_cells, n_cells])
+    grid_norm_unit = np.zeros([n_cells, n_cells])
 
-    grid_count_frames = np.zeros([n_cells, n_cells])
+    factor = np.float32(n_cells / max_width)
 
-    for ts in universe.trajectory:
+    for bead in selection:
```
richardjgowers, line 66: I don't think you need to (or should) loop over atoms.  Instead something like `np.rint(bead.positions[:, :2] / factor)` should give you an array of the `l` and `m` values.

membrane_curvature/lib/mods.py
```diff
@@ -24,66 +38,58 @@ def grid_map(coords, factor):
...
-    in each frame.
+    Returns set of z coordinates in grid.
 
     """
+    z_ref = np.zeros([n_cells, n_cells])
+    grid_z_coordinates = np.zeros([n_cells, n_cells])
+    grid_norm_unit = np.zeros([n_cells, n_cells])
 
-    grid_count_frames = np.zeros([n_cells, n_cells])
+    factor = np.float32(n_cells / max_width)
```
orbeckst, line 64: Is there a special reason for casting to np.float32? Why not just `n_cells / max_width` to keep it simple?

My reason is that you remove something that's mysterious not obvious why it needs to be there so that reading the code becomes easier. If at some point you _need_ an array top be float32, explicitly set the array's dtype.

membrane_curvature/lib/mods.py
```diff
@@ -2,19 +2,33 @@
 import numpy as np
 
 
+def get_positions(index):
```
orbeckst, line 5: I don't think this function should be needed. See if you can write your code with numpy arrays and broadcasting/slicing.

membrane_curvature/lib/mods.py
orbeckst: ANd you'll get rid of `get_positions()`.

membrane_curvature/lib/mods.py
```diff
@@ -24,66 +38,58 @@ def grid_map(coords, factor):
...
-            if grid_norm_unit[i, j] > 0:
-                z_Ref[i, j] += grid_z_coordinates[i, j] / grid_norm_unit[i, j]
-                grid_count_frames[i, j] += 1
 
-    for i, j in it.product(range(n_cells), range(n_cells)):
-        if grid_count_frames[i, j] > 0:
-            z_Ref[i, j] /= grid_count_frames[i, j]
+def avg_unit_cell(z_ref, n_cells, grid_z_coordinates, grid_norm_unit):
 
+    for i, j in it.product(range(n_cells), range(n_cells)):
```
orbeckst, line 86: See if you can eliminate this loop. Something like
```python
normed = grid_norm_unit > 0
z_ref[normed] += grid_z_coordinates / grid_norm_unit
z_ref[~normed] = np.nan
```

orbeckst (commented):
Small comments, mainly towards using numpy array operations. Sorry, did not manage to review everything in the time I had.

membrane_curvature/lib/mods.py
ojeda-e: Thanks for the comment @orbeckst Yes, there is a reason to use the `np.float32`. I discovered in the process of writing the code that the `/` operation is more expensive when I perform `n_cells / max_width` due to the type of the variables: `int` and `numpy.float64`, respectively. The `np.float32` was introduced to speed up. Since it is an operation that runs in every frame, the difference in performance is significant. To give you an estimate, in a system with only 500 lipid headgroups to define the surfaces per leaflet, the calculations runs 20% faster using the `np.float32` trick.

IAlibay: I think I'm mis-reading this, aren't `max_width` and `n_cells` both `ints` (going by both the docstring and the tests)?

Or are you referring to another division somewhere else?

IAlibay: @fiona-naughton just pointed out the following to me: https://github.com/MDAnalysis/membrane-curvature/blob/97713b39675940ea141b2980ae948db74fa3d5e5/membrane_curvature/core.py#L27

In any case, casting to np.float32 is more expensive for this specific operation on my end, does the performance difference maybe appear later on in the calculation?

ojeda-e: Thanks @IAlibay . In real-world systems the performance is significantly better using `np.float32`. The tests added so far were simplified as suggested by @lilyminium (See #22 and #31). I would prefer to keep it in the code, even if for the tests it looks like overkill. (I can remove it later if the benchmarks say otherwise)

membrane_curvature/lib/mods.py
ojeda-e: Thanks for your comment @richardjgowers. It also simplifies the function to the point `grid_map` isn't necessary. Since I also need the z coordinate as a float, I modified your suggestion to the following
`cell_xy_floor = np.int32( selection.positions[:,:2] / factor )`
`z_coordinate = selection.positions[:, 2] / factor * 10`.

membrane_curvature/lib/mods.py
ojeda-e: This is very cool, thanks @orbeckst. Your suggestion was included in the function `avg_unit_cell`. The previous tests work as expected.

ojeda-e:
After reviews, functions `get_positions` and `grid_map` were replaced by single lines in the function `derive_surface`. Two tests were added for this function. One using previously used small gro file, and passing coordinates as np.arrays in `mda.Universe`.

membrane_curvature/lib/mods.py
IAlibay: I still heavily recommend against early optimization, especially given that we don't seem to know what is actually benefitting from the cast? (it 100% can't be line 64)

Anyways, at the end of the day I don't see it as a big deal (as long as it doesn't impact accuracy), so I'll default to the other mentor's opinions here (especially given that @lilyminium's suggestion https://github.com/MDAnalysis/membrane-curvature/pull/40#discussion_r661054536 would change this code completely?).

However, I will add the following as things that probably need addressing:

1. Both `n_cells` and `max_width` are written in the docstring as `int` inputs. Is this meant to be the case? If not, the docstring needs to be updated accordingly.

2. Especially if they are meant to be `int`, then assuming we still want to keep this optimization in, why `np.float32`? Unless you're doing heavy vector operations, shouldn't a built-in `float` for a single value always be equal or faster in performance? (if they are `int` then it will default to a built-in `float` so it shouldn't even be necessary to cast?)

ojeda-e: Thanks for highlighting the docstrings, `max_width` is not `int` type, it's `float`. I'll update them. 
I don't see anything wrong with keeping it anyway. I'll leave and will re-evaluate in a performance test.

orbeckst:
If you leave the voodoo cast here, raise an issue to investigate further. From the amount of comments you got here you can see that nobody likes code that we don’t understand.

ojeda-e:
Thanks @orbeckst . Issue opened (#42), the line was changed to calculate with `factor = n_cells / max_width` only.

membrane_curvature/lib/mods.py
ojeda-e: > Hi @ojeda-e, I think I had something different in mind when I requested that you only use numpy etc and break it down to basics. Could you have a go at writing the function with the input arguments I suggested? I haven't had a look at tests because I assume they'll change with the new function.
> 
> @p-j-smith had a good suggestion with the scipy function. When I said "numpy only" that was probably too restrictive, sorry; totally on me. For future reference, you should consider any package safe if it's something that MDAnalysis considers a ["core dependency"](https://userguide.mdanalysis.org/stable/module_imports.html#list-of-core-module-dependencies). I actually find it preferable to use numpy/scipy functions because they're usually written by people much smarter than me, who've thought about the problem longer than me -- this usually results in faster, better code.
> 
> If you try something like the code below, you might make this a very short function!
> 
> ```python
> x, y, z = coordinates.T
> scipy.stats.binned_statistic_2d(x, y, z, statistic="mean", bins=..., range=...)
> ```

Hi @lilyminium, the PR has a lot of different feedback at this point and I ended up following the suggestion from @richardjgowers which conflicts with this one. Could we defer this to an issue as it would require me to add a new dependency and can be done in parallel with other work?

lilyminium: > def get_z_surface(coordinates, n_x_bins=10, n_y_bins=10, x_range=(0, 100), y_range=(0, 100)):

I don't think the feedback about the function signature contradicts @richardjgowers. You don't have to use scipy now if you don't want to -- the main gist is to have the main function work solely on numpy arrays of positions. I think you'll find this helpful for refactoring in #41; if you have a look at existing AnalysisBase classes, the important functions very rarely work on MDAnalysis Universes. Instead, the positions are extracted in `_single_frame` and these numpy arrays are used for the actual computation.

You've already done most of the work, to be honest. The simplest, most immediate conversion would be:

```python
def derive_surface(n_cells, selection, max_width):
	coordinates = selection.positions
	return get_z_surface(coordinates, n_x_bins=n_cells, n_y_bins=n_cells,
					     x_range=(0, max_width), y_range=(0, max_width))

def get_z_surface(coordinates, n_x_bins=10, n_y_bins=10, x_range=(0, 100), y_range=(0, 100)):
	z_ref = np.zeros((n_x_bins, n_y_bins))
	grid_z_coordinates = np.zeros((n_x_bins, n_y_bins))
    grid_norm_unit = np.zeros((n_x_bins, n_y_bins))

    x_factor = n_x_bins / (x_range[1] - x_range[0])
    y_factor = n_y_bins / (y_range[1] - y_range[0])
    x_coords, y_coords, z_coords = coordinates.T
    cell_x_floor = np.floor(x_coords * x_factor).astype(int)
    cell_y_floor = np.floor(y_coords * y_factor).astype(int)
    for l, m, z in zip(cell_x_floor, cell_y_floor, z_coords):
    	... # rest of current derive_surface
```

lilyminium: I also wouldn't be afraid of adding a new dependency (especially from the MDAnalysis core dependencies). It's mostly a matter of adding the package name to:
* https://github.com/MDAnalysis/membrane-curvature/blob/main/devtools/conda-envs/test_env.yaml
* https://github.com/MDAnalysis/membrane-curvature/blob/main/docs/requirements.yaml

ojeda-e: Thanks, I just thought this was going to be a separate issue. In the function you suggest the number of x and y bins is different, which I thought was going to be addressed later in #35 (together with PBC issue #36). 

Would you please confirm that is better to add this change here? @lilyminium 
Another option is to use your suggested function but without independent `x` and `y` (i.e. square arrays) and then complete the work when I address #35 later on?

lilyminium: IMO since you know that you want to address #35 in the future, there's no harm in designing your code towards that direction now. If you think it's easier to write and test a function that only takes `get_z_surface(coordinates, n_bins=10, range=(0, 100))` and then modify all the tests you write here for the `get_z_surface(coordinates, n_x_bins=10, n_y_bins=10, x_range=(0, 100), y_range=(0, 100))` signature later, go for it -- but to me it sounds like extra work.

If you have no intention of testing non-square arrays in this PR, you could just pass the same number into `n_x_bins` and `n_y_bins` as shown in the `derive_surface` example above. Then when you want to add rectangular functionality, you don't need to modify the existing tests from this PR.

ojeda-e: Thanks for the quick reply. I'll add it here then.

ojeda-e:
In this PR:

- [x]  Function `derive_surface` slightly modified from suggestion by @lilyminium to consider different max_width in `x` and `y` dimensions. i.e. `derive_surface(n_cells, selection, max_width_x, max_width_y)`. 
- [x] Smoke test added for rectangular and square grid using `pytest.mark.parametrize` for `get_z_surface`.
- [x] Previous tests modified accordingly to changes from `get_z_surface`.

Different binning for `x` and `y` will be addressed in upcoming issue #35

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -144,44 +151,152 @@ def test_mean_curvature_all():
...
+    np.array([[10., 20., 30.],
+              [10., 20., 30.],
+              [10., 20., 30.]]
+             ))])
+def test_avg_unit_cell_identity_other_values(n_cells, grid_z_coords):
+    z_ref = np.zeros([n_cells, n_cells])
+    unit = np.ones([n_cells, n_cells])
+    z_avg = avg_unit_cell(z_ref, grid_z_coords, unit)
+    for cell in range(n_cells):
+        assert_almost_equal(z_avg[cell], grid_z_coords[cell])
```
lilyminium, line 184: `assert_almost_equal` works for 2D arrays as well so you can get rid of the for loop altogether. You can also use `pytest.mark.parametrize` to bundle the two test cases together.

```suggestion
@pytest.mark.parametrize("n_cells, grid_z_coords", [
    (3, np.full((3, 3), 10)),
    (3, np.array([[10, 20, 30], [10, 20, 30], [10, 20, 30]], dtype=float)),
])
def test_avg_unit_cell_identity(n_cells, grid_z_coords):
    z_ref = np.zeros((n_cells, n_cells))  # this is not necessary if you go with the trick suggested above
    unit = np.ones((n_cells, n_cells))
    averaged_surface = avg_unit_cell(z_ref, grid_z_coords, unit)
    assert_almost_equal(averaged_surface, grid_z_coords)
```

membrane_curvature/lib/mods.py
```diff
@@ -1,89 +1,105 @@
...
 
-        for i, j in it.product(range(n_cells), range(n_cells)):
-            if grid_norm_unit[i, j] > 0:
-                z_Ref[i, j] += grid_z_coordinates[i, j] / grid_norm_unit[i, j]
-                grid_count_frames[i, j] += 1
 
-    for i, j in it.product(range(n_cells), range(n_cells)):
-        if grid_count_frames[i, j] > 0:
-            z_Ref[i, j] /= grid_count_frames[i, j]
+def avg_unit_cell(z_ref, grid_z_coordinates, grid_norm_unit):
```
lilyminium, line 78: Could you come up with a more descriptive name, like `average_z_coordinates`? I'm not really sure what averaging the unit cell means with respect to the arguments given to the function.

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -144,44 +151,152 @@ def test_mean_curvature_all():
...
+    # avg z coordinate in grid
+    np.array([[5., 10., 10.],
+              [10., 5., 10.],
+              [10., 10., 5.]]),
+)])
+def test_avg_unit_cell_more_beads(n_cells, grid_z_coords, grid_norm, grid_avg):
+    z_ref = np.zeros([n_cells, n_cells])
+    unit_cell = avg_unit_cell(z_ref, grid_z_coords, grid_norm)
+    for cell in range(n_cells):
+        assert_almost_equal(unit_cell[cell], grid_avg[cell])
```
lilyminium, line 207:
```suggestion
def test_avg_unit_cell_more_beads():
    grid_z_coords = np.full((3, 3), 10, dtype=float)
    grid_norm = np.array([[2, 1, 1],
                          [1, 2, 1],
                           1, 1, 2]], dtype=int)
    expected_surface = np.array([[5, 10, 10],
                                 [10, 5, 10],
                                 [10, 10, 5]], dtype=float)
    z_ref = np.zeros((n_cells, n_cells))
    averaged_surface = avg_unit_cell(z_ref, grid_z_coords, grid_norm)
    assert_almost_equal(averaged_surface, expected_surface)
```

You don't need to use `pytest.mark.parametrize` if you don't have multiple test cases to run through.

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -144,44 +151,152 @@ def test_mean_curvature_all():
...
+@pytest.mark.parametrize('n_cells, max_width, z_avg', [(
+    # number of cells
+    # max_width is 3nm/30A (from GRO_PO4_SMALL)
+    3, 30.,
+    # z ref of surface
+    # z coordinate in grid
+    np.array(([150., 150., 120.],
+              [150., 120., 120.],
+              [150., 120., 120.])))])
+def test_derive_surface(small_grofile, n_cells, max_width, z_avg):
```
lilyminium, line 219: Could you please rewrite this test to not use `parametrize` (since it does not use multiple test cases) and only with `assert_almost_equal` instead of the for loop?

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -144,44 +151,152 @@ def test_mean_curvature_all():
...
-    (0, 1), (-1, 1), (2, 1),
-    (0, 2), (-1, 2), (2, 2),
-    # should map to
-    (0, 0), (2, 0), (2, 0),
-    (0, 1), (2, 1), (2, 1),
-    (0, 2), (2, 2), (2, 2)])
-@pytest.mark.xfail(reason='Incorrect mapping of negative coordinates')
-def test_grid_map_9grid_negative(dummy_coords, expected):
-    assert grid_map(dummy_coords, 1) == expected
+@pytest.mark.parametrize('dummy_array', [(
```
lilyminium, line 287: This doesn't need to be parametrized either, as there's only one entry. Could you put `dummy_array` inside the test function?

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -144,44 +151,152 @@ def test_mean_curvature_all():
...
+              [200., 100., 150.],
+              [0., 200., 150.],
+              [100., 200., 150.],
+              [200., 200., 150.]]),
+    # z_avg in grid
+    np.array(([150., np.nan, 150.],
+              [np.nan, 150., 150.],
+              [150., 150., 150.]))
+)])
+def test_derive_surface_from_numpy_with_nans(dummy_array, n_cells, max_width, z_avg):
```
lilyminium, line 269: Could you please delete this test, and it's now just a duplicate of test_get_z_surface?

membrane_curvature/tests/test_mdakit_membcurv.py
```diff
@@ -144,44 +151,152 @@ def test_mean_curvature_all():
...
+              [200., 100., 120.],
+              [0., 200., 120.],
+              [100., 200., 120.],
+              [200., 200., 120.]]),
+    # z_avg in grid
+    np.array(([150., 150., 120.],
+              [150., 120., 120.],
+              [150., 120., 120.]))
+)])
+def test_derive_surface_from_numpy(dummy_array, n_cells, max_width, z_avg):
```
lilyminium, line 243: Could you please rewrite this test to use `get_z_surface` instead of `derive_surface`? This means you won't have to put the array into an MDAnalysis universe :)

lilyminium (changes requested):
Thanks for adding `get_z_surface` @ojeda-e. This more modular code is much more understandable to me! I have some comments on test -- mostly, `pytest.mark.parametrize` is not necessary if you only have one test case, and `assert_almost_equal` means you don't need for loops. I also have some tips for using numpy.

membrane_curvature/lib/mods.py
```diff
@@ -1,89 +1,105 @@
...
-    (x, y):  tuple
-        Value of (x, y) coordinates
-    factor:  float
-        Mappign factor to assign grid.
+    n_cells : int.
+        number of cells in the grid of size `max_width`.
+    selection : AtomGroup.
+        AtomGroup of reference selection to define the surface
+        of the membrane.
+    max_width : float.
```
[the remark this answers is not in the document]
ojeda-e: Thanks. I will  update to `derive_surface(n_cells_x, n_cells_y, selection, max_width_x, max_width_y)` (and respective tests if applicable).

membrane_curvature/lib/mods.py
```diff
@@ -1,89 +1,105 @@
...
+        Maximum width of simulation box. (Determined by simulation box dimensions)
 
-        Returns
-        -------
-        Returns l, m  with l,m as int.
+    Returns
+    -------
+    Returns set of z coordinates
```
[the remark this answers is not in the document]
ojeda-e: Yeah, thanks for this comment. I wasn't sure what to put here. :)

membrane_curvature/lib/mods.py
```diff
@@ -1,89 +1,105 @@
...
-    z_Ref : dict.
-        Dictionary to store values in every iteration over frames.
-    n_cells : int.
-        number of cells in the grid of size `max_width`.
-    selection : AtomGroup
-        AtomGroup of reference selection to define the surface
-        of the membrane.
-    max_width : int.
-        Maximum width of simulation box.
+    coordinates : tuple.
```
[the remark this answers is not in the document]
ojeda-e: Thanks. Let me confirm before submitting changes, @lilyminium. (In the process I clarify what I think is supposed to be correct). The structure of the parameters in the docstring is
```
parameter: type
    description
``` 
correct?
Is it common to add the shape of the numpy arrays in the type? Here I had it in the description. So in general, it should be like 
```
coordinates : numpy.ndarray 
     Coordinates of AtomGroup. Numpy of shape=(n_atoms, 3).
```
or should I specify the shape with the type as you suggest below for tuples?

membrane_curvature/lib/mods.py
ojeda-e: Sorry, missed this comment. Will be pushed shortly.

membrane_curvature/lib/mods.py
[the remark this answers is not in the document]
lilyminium:
```
parameter: type
    description
```

Erm, that's probably the correct one. I'll be honest that I'm not totally sure what's correct here. As long as the docs look good, either one works.

membrane_curvature/lib/mods.py
```diff
@@ -1,89 +1,105 @@
...
+        Numpy.ndarray with shape=(n_atoms, 3).
+    n_x_bins : int.
+        Number of bins in grid in the x dimension. 
+    n_y_bins : int.
+        Number of bins in grid in the y dimension. 
+    x_range : tuple
+        Range of indexes in grid in the x dimension with shape=(0, max_width_x).
```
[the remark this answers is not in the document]
lilyminium: I still think this would be the minimum x coordinate and the maximum x coordinate in angstrom, instead of the indexes in the grid. It's a replacement for `max_width` rather than n_cells, right?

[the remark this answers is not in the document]
ojeda-e: Sorry, I thought I included this change already. Give me a sec to update.

membrane_curvature/tests/test_mdakit_membcurv.py
ojeda-e: And I think I missed this too. Next push will include these changes.

lilyminium (approved):
Looks good to me!

Issue #21, opened 2021-06-16
Use explicit imports - change location of `lib/mods.py` 


ojeda-e:
As suggested in https://github.com/MDAnalysis/membrane-curvature/pull/13#discussion_r651377651, 
it's better to use explicit imports, which makes the code easier to debug and to read. The preferred form to use is `import y` and then use `y.x`.

To do:
- [ ] Change location of `libs/mods.py`
- [ ] Use explicit import in `core.py`.
- [ ] Use explicit import in `tests/test_mdakit_membcurv.py`

ojeda-e:
I have a question now that I review this issue. 
Would it make sense if I change the current structure of `membrane_curvature `
```
membrane_curvature
├── __init__.py
├── core.py
├── data
├── lib
│   └── mods.py
```

to 
```
membrane_curvature
├── __init__.py
├── core.py
├── data
├── analysis
│   └── Curvature.py. --> this would contain H, K
│   └── Surface.py. --> this would contain z_ref, derive_surface, normalized_grid
```
With `lib` and `mods`  gone, the explicit import would look like

```
from membrane_curvature.analysis.Curvature import mean_curvature, gaussian_curvature
from membrane_curvature.analysis.Surface import normalized_grid, derive_surface, get_z_surface
```
Otherwise, `from membrane_curvature.lib.mods import mean_curvature, gaussian_curvature, normalized_grid, derive_surface, get_z_surface` will solve this. 

Any suggestions @lilyminium?

lilyminium:
What do you envision being in `core.py`?

If you only have two files in `analysis`, you may as well have those files in the root level directory as well.

```
membrane_curvature
├── __init__.py
├── core.py
├── data
├── Curvature.py. --> this would contain H, K
├── Surface.py. --> this would contain z_ref, derive_surface, normalized_grid
```

One thing I will say is that it's very uncommon to capitalize module names.

ojeda-e:
Thanks for pointing out capitalization. I left `core.py` but that could be replaced by `base.py` (I don't see any functionality for `core.py` at the moment, but I may be wrong). I added the data folder but not tests etc.  so the tree would look like
```
membrane_curvature
├── __init__.py
├── base.py
├── curvature.py. --> this would contain H, K
├── surface.py. --> this would contain z_ref, derive_surface, normalized_grid
├── data
├── tests
```

lilyminium:
That looks like a good way to deal with it!

PR #43, opened 2021-07-06
AnalysisBase 

### LINKED ISSUE #41: AnalysisBase refactor
@ojeda-e could you please start putting your thoughts here on how you want the AnalysisBase refactor to go?

In general, AnalysisBase classes will have four methods defined:

* `__init__` this receives arguments. Usually an AtomGroup or Universe is provided here to operate over, and user-arguments such as `n_cell`.
* `_prepare` this typically sets up containers for the data to be collected in `_single_frame`, and maybe performs some pre-emptive computation.
* `_single_frame` this is the bulk of the class: the functionality that you want to operate over every frame of the trajectory. For example, you could be collecting lipid positions.
* `_conclude` this does the final computation with the data collected from `_single_frame`. Sometimes it's very short (e.g. https://github.com/MDAnalysis/mdanalysis/blob/64840d5df32ecb73ceadcc022dc4b71738537404/package/MDAnalysis/analysis/rms.py#L864-L870). Sometimes it's where the bulk of the computation is (e.g. https://github.com/MDAnalysis/mdanalysis/blob/64840d5df32ecb73ceadcc022dc4b71738537404/package/MDAnalysis/analysis/pca.py#L315-L322). For example, the calculation of mean and gaussian curvature could go here.
* (optional) post-run methods. If you don't want to compute the mean and gaussian curvature every time the user calls `run()` but only when they ask for it, you can add post-run methods instead, e.g. https://github.com/MDAnalysis/mdanalysis/blob/64840d5df32ecb73ceadcc022dc4b71738537404/package/MDAnalysis/analysis/pca.py#L431-L476

ojeda-e:
This PR is an initial suggestion to fix #41 

- Methods included in the initial analysis base:

- [x] `__init__` 
- [x] `_prepare`
- [x] `_single_frame`
- [x] `_conclude`

- Some additional comments

This first AnalysisBase was build under the following assumptions:
1. The atoms of reference in `AtomGroup` remain the same during the `n_frames` of the trajectory.
2. lipid translocation does not occur during the trajectory. (Although I am not considering bilayers, this reinforces 1)
3. By introducing a grid of bigger size than the simulation box, the calculation of `H` and `K` may be jeopardized by the introduction of `np.nans`. This selection will certainly generate `np.nans` in the calculation of the gradient, which ultimately will affect the averaged grids. Hence, the attempt of warning the user (in `__init__`). 
4. The overall result is the averaged `np.array` of `surface`, `K`, and `H` over frames. (in `_conclude`)
5. The calculation of `surface`, `H`, and `K`, is performed in every frame (in `_single_frame`).
6. The respective values of `surface`, `H` and `K` and their normalized grids are initialized (in `_prepare`). They are a tuple of np.arrays where the first element contains the respective values, and the second element counts the elements when defined.
7.  To sum and normalize values, the `_output` function was introduced. (But I am not sure this is something people would agree with)

As discussed in #21  with @lilyminium, this analysis base will potentially replace `core.py`.


membrane_curvature/base.py
```diff
@@ -0,0 +1,72 @@
...
+
+def _prepare(self):
+
+    # Initialize empty np.array(cumulative, count) of results
+    cumulative = np.zeros((self.x_bins, self.y_bins))
+    count = np.zeros((self.x_bins, self.y_bins))
+
+    self.results.surface = (cumulative, count)
+    self.results.mean = (cumulative, count)
+    self.results.gaussian = (cumulative, count)
```
lilyminium, line 34:
```suggestion
    self.results.surface = np.zeros((self.x_bins, self.y_bins))
    self.results.surface_count = np.zeros((self.x_bins, self.y_bins))
    self.results.mean = np.zeros((self.x_bins, self.y_bins))
    self.results.mean_count = np.zeros((self.x_bins, self.y_bins))
    self.results.gaussian = np.zeros((self.x_bins, self.y_bins))
    self.results.gaussian_count = np.zeros((self.x_bins, self.y_bins))
```
I think these results would be easier to work with if they were separate.

You also need to construct new arrays each time. As it is, currently, `self.results.gaussian[0] is self.results.mean[0]` will return True -- they are the same object in memory. That means, when you call `_output` on the surface, mean, and gaussian results, you are updating the same arrays 3 times.

membrane_curvature/base.py
```diff
@@ -0,0 +1,72 @@
...
+    # Calculate cumulative per frame when no nans.
+    cumulative_output += np.where(nans_in_grid, 0, value_per_frame)
+
+    # Count when no nans.
+    counter += np.where(nans_in_grid, 0, 1)
+
+
+def _single_frame(self):
+
+    surface_ = derive_surface(n_x_bins=self.n_cells_x, n_y_bins=self.n_cells_y,
```
lilyminium, line 53: Where do you give `derive_surface` the coordinates? I don't see `self.n_cells_x, self.n_cells_y, self.max_width_x, self.max_width_y` defined either?

membrane_curvature/base.py
```diff
@@ -0,0 +1,72 @@
...
+    # Save the results in NumPy array of shape (`n_x_bins`, `n_y_bins`).
+    self._output(surface_, self.results.surface[0], self.results.surface[1])
+    self._output(mean_, self.results.mean[0], self.results.mean[1])
+    self._output(gaussian_, self.results.gaussian[0], self.results.gaussian[1])
+
+
+def _conclude(self):
+
+    # Calculate average of np.array over trajectory normalized over counter
+    self.results.average_surface = self.results.surface[0] / self.results.surface[1]
```
lilyminium, line 68: Could you use `np.nanmean` to average over your surfaces without needing to have these `counter` arrays?

For example, you could set up a `z_surface` array to collect all the surfaces you get from `derive_surface` without modification:

```python
def _prepare(self):
    self.results.z_surface = np.full((self.n_frames, self.x_bins, self.y_bins), np.nan)

def _single_frame(self):
    self.results.z_surface[self._frame_index] = derive_surface(...)
```

and then finally average over the frames in _conclude:

```python
def _conclude(self):
   self.results.average_z_surface = np.nanmean(self.results.z_surface, axis=0)
```

(and the same for the mean and gaussian surfaces).

membrane_curvature/base.py
```diff
@@ -0,0 +1,72 @@
...
+             x_bins=100, y_bins=100,
+             x_range=None,
+             y_range=None,
+             pbc=True):
+    self.universe = universe
+    self.selection = select
+    self.pbc = pbc
+    self.y_bins = y_bins
+    self.x_bins = x_bins
+    self.x_range = x_range if x_range else (0, universe.dimensions[0])
```
lilyminium, line 17: Nice, I like these defaults!

membrane_curvature/base.py
```diff
@@ -0,0 +1,72 @@
...
+from membrane_curvature.lib.mods import derive_surface, mean_curvature, gaussian_curvature
+import numpy as np
+
+
+def __init__(self, universe, select,
+             x_bins=100, y_bins=100,
+             x_range=None,
+             y_range=None,
+             pbc=True):
+    self.universe = universe
```
lilyminium, line 12:
```suggestion
def __init__(self, universe, select,
             x_bins=100, y_bins=100,
             x_range=None,
             y_range=None,
             pbc=True, **kwargs):
    super().__init__(universe.trajectory, **kwargs)
    self.universe = universe
```

Calling the parent class `__init__` makes sure the base class instantiation is done as well!

membrane_curvature/base.py
```diff
@@ -0,0 +1,72 @@
+import numpy as np
+
+from membrane_curvature.lib.mods import derive_surface, mean_curvature, gaussian_curvature
+import numpy as np
+
+
+def __init__(self, universe, select,
```
lilyminium, line 7: Do you need `universe` to get passed? Is `select` an AtomGroup or a string -- if it's an AtomGroup, you could just use `universe = select.universe`.

FYI most AnalysisBase classes use `select` as a string, and many accept either an AtomGroup or Universe for the first argument. To safeguard against the multiple possibilities, you could do this:

```python
def __init__(self, universe, select="all", ..., **kwargs):
  super().__init__(universe.universe.trajectory, **kwargs)
  self.selection = universe.select_atoms(select)
```

membrane_curvature/base.py
```diff
@@ -0,0 +1,72 @@
...
+             x_bins=100, y_bins=100,
+             x_range=None,
+             y_range=None,
+             pbc=True):
+    self.universe = universe
+    self.selection = select
+    self.pbc = pbc
+    self.y_bins = y_bins
```
lilyminium, line 15: Could you please name this something more accurate, such as `n_y_bins`? `y_bins` could be the actual bins, i.e. `[(0, 10), (10, 20), (20, 30), ...]`

membrane_curvature/base.py
```diff
@@ -0,0 +1,72 @@
+import numpy as np
+
+from membrane_curvature.lib.mods import derive_surface, mean_curvature, gaussian_curvature
+import numpy as np
+
+
```
lilyminium, line 6:
```suggestion
from MDAnalysis.analysis.base import AnalysisBase

class MembraneCurvature(AnalysisBase):
```
I think you're missing an actual class definition!

lilyminium (changes requested):
This is a fantastic start @ojeda-e! Your intuition is spot on for what should be done in the init/prepare/single_frame/conclude methods. I have some notes on the structure, and I wonder if we can avoid using counter arrays entirely by using `np.nanmean`.

membrane_curvature/base.py
ojeda-e: Actually, I was about to edit when I received the review 😄. 

I had it initially in the explicit form as you suggested, but I found it ugly (long and redundant). Then I changed to the submitted version, but yes, that's incorrect. I will submit a change with the explicit version. 

What do you think about this option?

```
cumulative = np.zeros((self.x_bins, self.y_bins))
count = np.zeros((self.x_bins, self.y_bins))

self.results.surface = (cumulative.copy() , count.copy() )
self.results.mean = (cumulative.copy(), count.copy() )
self.results.gaussian = (cumulative.copy(), count.copy())
```

lilyminium: I still wouldn't be keen on having a tuple of arrays in the results. Simpler is often better, even if it results in more lines of code.

ojeda-e: Thanks for the comment @lilyminium. Maybe it's something silly, but may I ask why tuples of arrays are not the best approach here?

lilyminium: I'm mostly thinking of the user experience when I say this. If I wanted to work with the raw `mean` array, it's much handier for me to work with `self.results.mean_array` than `self.results.mean[0]`. The `[0]` might look like a small addition but it means that I have to remember which order the arrays are in (because it might be `self.results.mean[1]`), so I have to look up the documentation each time (in fact this is a real downside of the existing RMSD class in MDAnalysis -- the actual RMSD results are only the last three columns of the `self.results.rmsd` array).

It's also easier to document `self.results.mean_array is the array of the mean surfaces` instead of `self.results.mean is a tuple of arrays, the first array is the array of mean surfaces, the second is the array of counts` :)

Anyway, I think you may be able to get rid of the counter arrays by using `np.nanmean`.

ojeda-e: I see. Thanks for the clear explanation. 
The question was more out of curiosity to see the downside of the approach I used here. And to be honest, while writing code I never think of that, but as an user I can totally see it :) Very helpful, thanks again.

lilyminium: It was a good question! There's always a balance between making life easier for the developer but thinking about users too :p

membrane_curvature/base.py
```diff
@@ -0,0 +1,72 @@
...
+    # Initialize empty np.array(cumulative, count) of results
+    cumulative = np.zeros((self.x_bins, self.y_bins))
+    count = np.zeros((self.x_bins, self.y_bins))
+
+    self.results.surface = (cumulative, count)
+    self.results.mean = (cumulative, count)
+    self.results.gaussian = (cumulative, count)
+
+
+def _output(self, value_per_frame, cumulative_output, counter):
```
ojeda-e, line 37: I wanted to ask you something about this approach, @lilyminium . To be honest, I didn't see any other sensible option to accumulate the values per frame and count. Would you please share with me some alternatives? What do you think would be a good idea? 
(Not that I am looking for a change ☺️ , more to consider alternatives I can't see by myself)

lilyminium: Do you need to accumulate the count per frame as an end goal? My first thought would be to collect all the surfaces in a 3D array: (n_frames, n_x_cells, n_y_cells) and then average over them with `np.nanmean` for the average surface.

ojeda-e:
Hi @lilyminium 
I updated the `AnalysisBase` with the suggested changes and other minor improvements and corrections.
I also changed the position of the argument selection in the `derive_surface` function just for consistency.

lilyminium:
This design is looking really, really good @ojeda-e. I don't have much to say except that some lines are looking quite long so you may want to check that you're following PEP8. One trick could be to make `get_z_surface` a method in the class, since it uses attributes of `self` so often -- but that's minor.

I guess the proof is in the pudding -- would you like to start writing tests to see how it works? :-)

lilyminium:
Also great idea to switch the order of the arguments to `derive_surface`, the new order feels much more intuitive!

ojeda-e:
> I guess the proof is in the pudding -- would you like to start writing tests to see how it works? :-)

Yes! :D I will open an issue for the respective tests and will work on them after solving `pbc` conditions in `surface.py`. 

If you are ok with it, I will skip making `get_z_surface` as a method. At least temporarily. :) 
I will push the final version briefly.

ojeda-e:
@lilyminium re-reading your comment makes me wonder if you would prefer the tests in this PR or in a separate issue. Apologies I wrongly assumed it was a separate PR. What do you think works best?

lilyminium:
I would personally prefer in this PR, just so we make sure that everything in the main branch always works :-)

devtools/conda-envs/test_env.yaml
```diff
@@ -6,10 +6,10 @@ dependencies:
   - python
   - pip
   - numpy
-  - MDAnalysis
-  - mdtraj # temporary; remove after refactoring
   - sphinx
-
+  - MDAnalysis==2.0.0-dev0
```
RMeli, line 10: I think you need to install it explicitly via pip:

```suggestion
  - pip:
      - MDAnalysis==2.0.0-dev0
```

lilyminium:
Another catch is that I think it’s actually called MDAnalysis==2.0.0b0

RMeli:
@lilyminium I had this feeling and I was looking for the announcement mentioning the beta package name... Do you know where I can find it (for future reference)? It is not with the usual blogs. Was it just on Twitter?

IAlibay:
> @lilyminium I had this feeling and I was looking for the announcement mentioning the beta package name... Do you know where I can find it (for future reference)? It is not with the usual blogs. Was it just on Twitter?

We never got around to it, we were going to announce it with the MDAnalysis workshop materials but unfortunately that got stalled.

lilyminium:
Sorry I didn't mean to remove those review requests... hazards of githubbing on phone 🤦‍♀️

devtools/conda-envs/test_env.yaml
IAlibay: Because it's relying on the `Results` class, could you raise an issue as a reminder that MDA should get pinned to >=2.0.0 in `setup.py`? (I wouldn't do it now, but it'd be a good thing to do before the first release)

IAlibay: Also +1 to @RMeli's comment. The beta needs to be installed via `pip` (swapping `-dev0` with `b0` should do the trick).

ojeda-e:
Thanks, @lilyminium, @IAlibay, and @RMeli for your comments. 

I just wanted to add an initial version of the tests with previously used `dummy_arrays` and a Universe with `.gro` and `.xtc` files. They passed, thanks again!

membrane_curvature/base.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class MembraneCurvature(AnalysisBase):
+
+    def __init__(self, universe, select='all',
+                 n_x_bins=100, n_y_bins=100,
+                 x_range=None,
+                 y_range=None,
+                 pbc=True, **kwargs):
+
```
IAlibay, line 18: Probably already on your to-do list, but some docstring (especially detailing the input variables), would be very useful.

membrane_curvature/base.py
```diff
@@ -0,0 +1,68 @@
...
+        super().__init__(universe.universe.trajectory, **kwargs)
+        self.selection = universe.select_atoms(select)
```
IAlibay, line 20: Just to make everything clearer to future developers I would rename this from self.selection to either self.ag or self.atomgroup. That way folks can look at the name and immediately know what type of object it is (also better matches other MDA analysis methods).

membrane_curvature/base.py
```diff
@@ -0,0 +1,68 @@
...
+        self.pbc = pbc
+        self.n_x_bins = n_x_bins
+        self.n_y_bins = n_y_bins
+        self.x_range = x_range if x_range else (0, universe.dimensions[0])
+        self.y_range = y_range if y_range else (0, universe.dimensions[1])
+
+        # Raise if selection doesn't exist
+        if (self.selection.n_residues) == 0:
```
IAlibay, line 28: Nothing wrong with this technically, but it might be good to simplify it to just `len(atomgroup) == 0` purely on the the basis that it's more descriptive of your intent / more widely used way of doing this?

IAlibay (changes requested):
Just a couple of comments, overall looks good mainly needs documentation.

membrane_curvature/base.py
ojeda-e: Thanks, @IAlibay, I wasn't sure if it was going to be better/easier to open an issue after this PR was accepted. I will include docstrings in the next push.

IAlibay: At least for this first PR I'd suggest adding input/output parameters and anything that's raised. That way it's easier for everyone to keep track of what is intended to be what.

membrane_curvature/base.py
```diff
@@ -0,0 +1,68 @@
...
+            raise ValueError("Invalid selection. Empty AtomGroup.")
+
+        # Raise if range doesn't cover entire dimensions of simulation box
```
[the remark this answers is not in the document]
ojeda-e: I think that's something I didn't pay too much attention to (and I didn't clarify, my bad). One of the assumptions I had is that calculation of membrane curvature is performed in equilibrated systems.  If the box shrinks or expands too much (one unit cell in the grid), the calculation wouldn't be accurate. 

The reason why I introduced the restrictions (which I changed now to a warning message) is that (1) if the surface is derived from a portion of the membrane, the calculation of curvature in the edges will be inaccurate because the calculation of a  gradient runs over adjacent values. On the other hand, (2) in the scenario in which the grid is too big, unnecessary `np.nans` will be introduced. It means that, again, the undefined adjacent values will have a negative impact on the calculation of curvature. 

I don't see a straightforward solution for NPT systems unless the calculation returns a different grid per frame and then results must be normalized over grids of the same size, and I don't have a clear idea, at least right now, of how to keep track of such variation over frames. Do you have any suggestions here?

membrane_curvature/base.py
ojeda-e: By renaming `selection` to `ag`, your suggestion reads better, so I updated it. Thanks.

membrane_curvature/base.py
```diff
@@ -0,0 +1,126 @@
...
+        # Raise if selection doesn't exist
+        if len(self.ag) == 0:
+            raise ValueError("Invalid selection. Empty AtomGroup.")
+
+        # Warning message if range doesn't cover entire dimensions of simulation box
+        for dim_string, dim_range, num in [('x', self.x_range, 0), ('y', self.y_range, 1)]:
+            if (dim_range[1] < universe.dimensions[num]):
+                warnings.warn(f"Grid range in {dim_string} do not cover entire "
+                              "dimensions of simulation box.\n Minimum dimensions "
+                              "must be equal to simulation box.")
```
ojeda-e, line 92: @IAlibay this is the modified warning message, now updated for `x` and `y` dimensions.

ojeda-e:
As suggested by @IAlibay, docstrings were added and some suggested changes that improve the readability of the code. :)

lilyminium:
Hi Estefania,

I believe the DensityAnalysis in MDAnalysis deals with a similar problem. Perhaps have a look at how the box is defined there? If I recall correctly, the box doesn’t change but you may get warnings. ***@***.*** would know more as the one who wrote it!)

https://docs.mdanalysis.org/1.1.1/documentation_pages/analysis/density.html

https://docs.mdanalysis.org/1.1.1/_modules/MDAnalysis/analysis/density.html#DensityAnalysis

Cheers,
Lily

ojeda-e:
This PR includes the AnalysisBase with respective docstrings and tests. 
To limit the extent of this PR, the limitations pointed out by @IAlibay of possible NPT boxes will be in issue #47. PBC conditions will be addressed in issue #36 

I would appreciate a final review here @IAlibay @orbeckst @fiona-naughton

membrane_curvature/base.py
```diff
@@ -0,0 +1,126 @@
...
+:Copyright: GNU Public License v3
+
+MembraneCurvature calculates the mean and Gaussian curvature of
+surfaces derived from a selection of reference.
+"""
+
+import numpy as np
+import warnings
+from membrane_curvature.surface import get_z_surface
+from membrane_curvature.curvature import mean_curvature, gaussian_curvature
```
orbeckst, line 16: Use relative imports
```suggestion
from .surface import get_z_surface
from .curvature import mean_curvature, gaussian_curvature
```
to be explicit about the relationship between the modules. This makes clear that they are all in the same package.

membrane_curvature/base.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+from MDAnalysis.analysis.base import AnalysisBase
+import MDAnalysis as mda
```
orbeckst, line 20: I'd move the general `import MDAnalysis` before the specific one for `AnalysisBase` — it's no big thing but conceptually I like loading the foundation first so if any errors occur then I know that they are in the foundation and not the specifics. (MDAnalysis.analysis is not loaded when you do `import MDAnalysis` so it could conceivably help you in pinning down issues.)

membrane_curvature/base.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+class MembraneCurvature(AnalysisBase):
+
```
orbeckst, line 24: I _think_ in our style we don't have an empty line here.

membrane_curvature/base.py
```diff
@@ -0,0 +1,126 @@
...
+    select : str or iterable of str, optional. Default: 'all'
+        The selection string of an atom selection to use as a
+        reference to derive a surface.
+    pbc : bool, optional. Default: 'True'
+        Apply periodic boundary conditions.
+    n_x_bins : int, optional. Deafult: '100'
+        Number of bins in grid in the `x` dimension.
+    n_y_bins : int, optional. Default: '100'
+        Number of bins in grid in the `y` dimension.
+    x_range : tuple of (float, float), optional. Deafult: (0, `universe.dimensions[0]`)
```
orbeckst, line 41:
```suggestion
    x_range : tuple of (float, float), optional. Default: (0, `universe.dimensions[0]`)
```

membrane_curvature/base.py
```diff
@@ -0,0 +1,126 @@
...
+        Range of coordinates (min, max) in the `x` dimension.
+    y_range : tuple of (float, float), optional. Deafult: (0, `universe.dimensions[1]`)
```
orbeckst, line 43:
```suggestion
    y_range : tuple of (float, float), optional. Default: (0, `universe.dimensions[1]`)
```

membrane_curvature/base.py
```diff
@@ -0,0 +1,126 @@
...
+class MembraneCurvature(AnalysisBase):
+
+    r"""
+    MembraneCurvature is a tool to calculate membrane curvature.
+
+    Parameters
+    ----------
+    universe : Universe or AtomGroup
+        An MDAnalysis Universe object.
+    select : str or iterable of str, optional. Default: 'all'
```
orbeckst, line 32: The syntax for default values seems to be according to  [numpy docstring standard](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) (which is what we follow  [Userguide: Working with the docs](https://userguide.mdanalysis.org/2.0.0-dev0/contributing_code.html#working-with-the-code-documentation))

```suggestion
    select : str or iterable of str, optional, default: 'all'
```

but you can leave out the value because it's already part of the function signature.

membrane_curvature/base.py
```diff
@@ -0,0 +1,126 @@
...
+        The selection string of an atom selection to use as a
+        reference to derive a surface.
+    pbc : bool, optional. Default: 'True'
+        Apply periodic boundary conditions.
+    n_x_bins : int, optional. Deafult: '100'
```
orbeckst, line 37: default

membrane_curvature/base.py
```diff
@@ -0,0 +1,126 @@
...
+        Number of bins in grid in the `x` dimension.
```
orbeckst, line 38: We use single backquotes when referring to a parameter. For anything else we either use explicit reST roles (like :func: or :class: or :attr:) or use "inline code" _double_ backquotes
```suggestion
        Number of bins in grid in the ``x`` dimension.
```
ALternatively, you can also use the :math: role
```reST
        Number of bins in grid in the :math:`x` dimension.
```
although in this specific case, this seems less readable and you could also just write
```reST
        Number of bins in grid in the x dimension.
```

The main point is that when you mark-up something then think about why you are marking it up and what _meaning_ it should carry.

membrane_curvature/base.py
```diff
@@ -0,0 +1,126 @@
...
+        Arrays of shape (`n_frames`, `n_x_bins`, `n_y_bins`)
+
+    Returns
+    -------
+    average_z_surface : ndarray
+        Average of the array elements in `z_surface`
+    average_mean_curvature :
+        Average of the array elements in `mean_curvature`
+    average_gaussian_curvature:
+        Average of the array elements in `gaussian_curvature`
```
orbeckst, line 65: It does not make sense to specify return values for a class unless the class is pretending to be a function (with `__call__`) but that's not what Analysis classes do. They store results in `results` after executing `run()` — exactly as you documented. The `__init__` method does not have a return value.

Remove the *Returns* section. Add whatever `_conclude()` adds to `results` to the attribute section. Add some text explaining that these specific attributes are available after run.

membrane_curvature/base.py
```diff
@@ -0,0 +1,126 @@
...
+        self.n_x_bins = n_x_bins
+        self.n_y_bins = n_y_bins
+        self.x_range = x_range if x_range else (0, universe.dimensions[0])
+        self.y_range = y_range if y_range else (0, universe.dimensions[1])
+
+        # Raise if selection doesn't exist
+        if len(self.ag) == 0:
+            raise ValueError("Invalid selection. Empty AtomGroup.")
+
+        # Warning message if range doesn't cover entire dimensions of simulation box
```
orbeckst, line 87: add a note for later that you're only checking the first frame and that NPT simulations are not properly covered here

membrane_curvature/base.py
```diff
@@ -0,0 +1,126 @@
...
+        for dim_string, dim_range, num in [('x', self.x_range, 0), ('y', self.y_range, 1)]:
+            if (dim_range[1] < universe.dimensions[num]):
+                warnings.warn(f"Grid range in {dim_string} do not cover entire "
```
orbeckst, line 90: It's good practice to also log a warning using the [logging facility](https://docs.python.org/3/library/logging.html). See density analysis https://github.com/MDAnalysis/mdanalysis/blob/8f9f402fe65d6a90e88b60b6da735c130425f00e/package/MDAnalysis/analysis/density.py#L181-L183 and eg https://github.com/MDAnalysis/mdanalysis/blob/8f9f402fe65d6a90e88b60b6da735c130425f00e/package/MDAnalysis/analysis/density.py#L420-L424 as examples.

membrane_curvature/base.py
```diff
@@ -0,0 +1,126 @@
...
+    def _single_frame(self):
+        # Populate a slice with np.arrays of surface, mean, and gaussian per frame
+        self.results.z_surface[self._frame_index] = get_z_surface(self.ag.positions,
+                                                                  n_x_bins=self.n_x_bins,
+                                                                  n_y_bins=self.n_y_bins,
+                                                                  x_range=self.x_range,
+                                                                  y_range=self.y_range)
+
+        self.results.mean[self._frame_index] = mean_curvature(self.results.z_surface[self._frame_index])
+
```
orbeckst, line 117: I would eliminate the blank lines to make it easier to read it as one computing block but that's personal style.

membrane_curvature/base.py
```diff
@@ -0,0 +1,126 @@
...
+        self.results.gaussian[self._frame_index] = gaussian_curvature(self.results.z_surface[self._frame_index])
+
+    def _conclude(self):
+        # Calculate mean of np.array of surface, mean, gaussian
+        self.results.average_z_surface = np.nanmean(self.results.z_surface, axis=0)
+
+        self.results.average_mean = np.nanmean(self.results.mean, axis=0)
+
```
orbeckst, line 125: I would eliminate the blank lines to make it easier to read it as one computing block but that's personal style.

membrane_curvature/surface.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
 
 
-def derive_surface(n_cells_x, n_cells_y, selection, max_width_x, max_width_y):
+def derive_surface(selection, n_cells_x, n_cells_y, max_width_x, max_width_y):
```
orbeckst, line 4: - update the doc string
- renaming `selection` -> `atoms` would be clearer

membrane_curvature/tests/test_membrane_curvature.py
```diff
@@ -205,3 +206,84 @@ def test_get_z_surface(x_bin, y_bin, x_range, y_range, expected_surface):
...
+        return u
+
+    def test_invalid_selection(self, universe):
+        with pytest.raises(ValueError, match=r'Invalid selection'):
+            MembraneCurvature(universe, select='name P')
+
+    def test_grid_bigger_than_simulation_box(self, universe):
+        with pytest.warns(UserWarning):
+            MembraneCurvature(universe, select='name PO4', x_range=(0, 10))
+            MembraneCurvature(universe, select='name PO4', y_range=(0, 10))
```
orbeckst, line 234: You need to test these two lines separately, otherwise you can get a passing test even if one of them does not raise the warning.

orbeckst: `pytest.warns` can also match part of the warnings string, always a good idea to test that the warning looks like what you expect it.

orbeckst (changes requested):
Your AnalysisBase version looks very good — clean structure, good use of functions, well done!

I have a bunch of comments on documentation and testing. In addition to the inline comments I noted that your tests always set explicitly `selection="all"`. You should also test that the implicit default works (just in case this gets changed) and you should also check that other selections work.


Again: well done!

devtools/conda-envs/test_env.yaml
orbeckst: Btw, @IAlibay, this PR demonstrates to be that the beta was worth the effort because it makes it a lot easier for packages to test against what 2.0 will look like.

membrane_curvature/base.py
```diff
@@ -54,15 +56,42 @@ class MembraneCurvature(AnalysisBase):
...
+        Average of the array elements in `gaussian_curvature`.
+        Each array has shape (`n_x_bins`, `n_y_bins`)
+
+
+    Notes
+    -----
+    The derived surface and calculated curvatures are available in the
+    ``.results``::atttributes. 
+
+    The attribute :attr:`MembraneCurvature.z_surface`
```
orbeckst, line 75: Isn't this inside `results`? So
```suggestion
    The attribute :attr:`MembraneCurvature.results.z_surface`
```

If you only want to display `z_surface` use the tilde
```reST
    The attribute :attr:`~MembraneCurvature.results.z_surface`
```

membrane_curvature/base.py
```diff
@@ -54,15 +56,42 @@ class MembraneCurvature(AnalysisBase):
...
+    contains the derived surface averaged over the `n_frames` of the trajectory.
+
+    The attributes :attr:`MembraneCurvature.mean_curvature` and
```
orbeckst, line 78: results?

membrane_curvature/base.py
```diff
@@ -54,15 +56,42 @@ class MembraneCurvature(AnalysisBase):
...
+    :attr:`MembraneCurvature.gaussian_curvature` contain the computed values of
```
orbeckst, line 79: results?

membrane_curvature/base.py
```diff
@@ -54,15 +56,42 @@ class MembraneCurvature(AnalysisBase):
...
-    -------
-    average_z_surface : ndarray
-        Average of the array elements in `z_surface`
-    average_mean_curvature :
-        Average of the array elements in `mean_curvature`
-    average_gaussian_curvature:
-        Average of the array elements in `gaussian_curvature`
+    Example
+    -----------
+    You can pass a universe containing your selection of reference:
```
orbeckst, line 84: Needs double colon to format as code
```suggestion
    You can pass a universe containing your selection of reference::
```

membrane_curvature/base.py
```diff
@@ -54,15 +56,42 @@ class MembraneCurvature(AnalysisBase):
...
+
+        import MDAnalysis as mda
+        from MDAnalysis.analysis import MembraneCurvature
+
+        u = mda.Universe(coordinates, trajectory)
+        mc = MembraneCurvature(u).run()
+
+        surface =  mc.results.average_z_surface
+        mean_curvature =  mc.results.average_mean_curvature
+        gaussian_curvature = mc.results.average_gaussian_curvature
```
orbeckst, line 94: Maybe two lines for how to plot with matplotlib to round it off?

membrane_curvature/base.py
```diff
@@ -15,33 +15,35 @@
...
 from membrane_curvature.curvature import mean_curvature, gaussian_curvature
 
-
-from MDAnalysis.analysis.base import AnalysisBase
 import MDAnalysis as mda
+from MDAnalysis.analysis.base import AnalysisBase
 
+import logging
+
+logger = logging.getLogger("MDAnalysis.analysis.density")
```
orbeckst, line 23: name your logger differently ... something like
```suggestion
logger = logging.getLogger("MDAnalysis.MDAKit.membrane_curvature")
```
I started your logger's name with "MDAnalysis" so that output gets added to the default logger that can be started with
```python
import MDAnalysis
MDAnalysis.start_logging()
```
and whose root logger is named "MDAnalysis". Hence any logger named "MDAnalysis.*" will also log under it.

I suggest "MDAnalysis.MDAKit" to define a namespace that all MDAKits could be using but we haven't really thought about this so this could definitely be subject to change. Maybe @IAlibay @lilyminium @richardjgowers @fiona-naughton have opinions.

In any case, for right now, pretty much anything is ok that's not "density" ;-).

orbeckst (approved):
Looking fantastic. There's one required fix with the logger but I am sure you'll come up with a sensible solution there so I'll just approve.

membrane_curvature/base.py
ojeda-e: Added.

orbeckst:
That was an important merged PR, well done!

PR #49, opened 2021-07-17
MDA>=2.0.0 version added to `install_requires`

### LINKED ISSUE #46: Version of MDA in `setup.py`
@ojeda-e 
Because it's relying on the `Results` class, could you raise an issue as a reminder that MDA should get pinned to >=2.0.0 in `setup.py`? (I wouldn't do it now, but it'd be a good thing to do before the first release)

_Originally posted by @IAlibay in https://github.com/MDAnalysis/membrane-curvature/pull/43#discussion_r667237549_

ojeda-e:
This PR fixes #46 

- [x] In `setup.py`, three packages were added to install requires:
- 'numpy>=1.20.2',
- 'mdanalysis>=2.0.0',
- 'mdanalysistests>=2.0.0'.

- [x] For NumPy:
- Check if installed.
- Check `Python>=3.6`

@IAlibay, would you please confirm no more changes are needed here at the moment?


setup.py
```diff
@@ -42,7 +63,7 @@
     include_package_data=True,
 
     # Allows `setup.py test` to work correctly with pytest
-    setup_requires=[] + pytest_runner,
+    setup_requires=['numpy>=1.20.2'] + pytest_runner,
```
tylerjereddy, line 66: That's a fairly new version of NumPy--are you leveraging some bleeding edge features for the analysis work?

setup.py
```diff
@@ -12,12 +12,33 @@
 needs_pytest = {'pytest', 'test', 'ptr'}.intersection(sys.argv)
 pytest_runner = ['pytest-runner'] if needs_pytest else []
 
+# Check Python version
+if sys.version_info[:2] < (3, 6):
+    print('This package requires a verion of Python>=3.6.'
+          'Please upgrading your version of Python to use MembraneCurvature.')
+    sys.exit(-1)
```
tylerjereddy, line 19: might be a little more conventional to `raise RuntimeError(message)`

IAlibay: I suspect @ojeda-e implemented this because it's the way we do it in MDAnalysis core. That being said, we probably should change that too (I just saw some py2 stuff in there too that needs to get deleted =.=).

setup.py
orbeckst: As a general rule, be as lenient about your required versions as possible. It can be very difficult for a package manager to resolve specific version requirements, especially on slightly older installations with very new package versions.

setup.py
```diff
@@ -12,12 +12,33 @@
...
+
+# Add release
+RELEASE = "0.0.0-dev0"
+
+# Check numpy is installed
 try:
+    import numpy as np
+except ImportError:
+    print("MembraneCurvature requires numpy>=1.20.2")
```
orbeckst, line 28: I would not include the version requirement here, just keep information in _one_ place (DRY = don't repeat yourself). It's enough to fail here if you don't have numpy. Versions are handled by the package manager.

setup.py
```diff
@@ -12,12 +12,33 @@
...
+
     with open("README.md", "r") as handle:
         long_description = handle.read()
-except:
-    long_description = "\n".join(short_description[2:])
 
+    # set requirements
+    install_requires = [
+        'numpy>=1.20.2',
+        'mdanalysis>=2.0.0',
```
orbeckst, line 39: Do these requirements include 2.0.0b, or do you want to exclude 2.0.0b? Look at how version information is resolved by setuptools.

setup.py
```diff
@@ -12,12 +12,33 @@
...
+if __name__ == "__main__":
+
     with open("README.md", "r") as handle:
         long_description = handle.read()
-except:
-    long_description = "\n".join(short_description[2:])
 
+    # set requirements
+    install_requires = [
+        'numpy>=1.20.2',
```
orbeckst, line 38: numpy 1.20.2 really required?

setup.py
ojeda-e: Thanks for this. I wrongly assumed the NumPy version required should match the one used in the environment the tool is developed. In this case, `numpy>=1.16` will do.

setup.py
ojeda-e: Yes, you're correct @IAlibay. I will follow @tylerjereddy suggestion here.

setup.py
ojeda-e: Not, not really. Changed to 1.16.0. :)

setup.py
ojeda-e: It's actually the `2.0.0b0` version. Updated in the last push.

ojeda-e:
Thanks for your comments @tylerjereddy and @orbeckst.
Although tests passed with `Python>=3.7`, I added `Python=3.6` and `numpy>=1.16.0` so the requirements will be consistent with those in MDAnalysis.

setup.py
```diff
@@ -12,12 +12,31 @@
 needs_pytest = {'pytest', 'test', 'ptr'}.intersection(sys.argv)
 pytest_runner = ['pytest-runner'] if needs_pytest else []
 
+# Check Python version
+if sys.version_info[:2] < (3, 6):
+    raise RuntimeError('Python version >=3.6 required.')
+
+# Add release
+RELEASE = "0.0.0-dev0"
```
lilyminium, line 20: I think this is probably not required -- MDAnalysis core uses it to

* set up the version in the final `setup()` call
* specify which version of the tests is required
* check if it's a release version (no "dev") or development version.

You probably don't need any of those unless you plan to manage concurrent released and development versions the way MDAnalysis does, with its own docs and C/++ extensions. As you are using ReadTheDocs for documentation currently, RTD manages your doc versioning for you.

setup.py
```diff
@@ -12,12 +12,31 @@
...
+
+# Check numpy is installed
 try:
+    import numpy as np
+except ImportError:
+    print("MembraneCurvature requires NumPy.")
```
lilyminium, line 26: @tylerjereddy @IAlibay @orbeckst I forget why MDAnalysis has all this numpy complexity -- something to do with setuptools not finding the right `include` directory for compilation? Do you think @ojeda-e can just skip this check and leave it in `install_requires` if she's not compiling any extensions?

Regardless, I would leave this as an `ImportError`, possibly with a more informative error message. At this point `setup.py` just prints the message and keeps going. The original MDAnalysis setup.py has `sys.exit(-1)` [afterwards which quits Python with an error code](https://github.com/MDAnalysis/mdanalysis/blob/8f9f402fe65d6a90e88b60b6da735c130425f00e/package/setup.py#L195), but IMO this is what the error is for.

.github/workflows/CI.yaml
```diff
@@ -25,7 +25,15 @@ jobs:
       fail-fast: false
       matrix:
         os: [macOS-latest, ubuntu-latest, windows-latest]
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9]
+        include:
+          - name: numpy_min
+            os: ubuntu-latest
+            python-version: 3.6
+            run_type: FULL
```
lilyminium, line 33: What is `FULL`? I don't think you use this line

.github/workflows/CI.yaml
```diff
@@ -25,7 +25,15 @@ jobs:
...
+            install_hole: false
```
lilyminium, line 34: You definitely don't need to install HOLE

.github/workflows/CI.yaml
```diff
@@ -25,7 +25,15 @@ jobs:
...
+            codecov: false
+            numpy: numpy=1.16.0
```
lilyminium, line 36: Have you checked this test to see if it's actually installing 1.16.0 (and paying attention to the `codecov` and `python-version` flags? My hunch is not necessarily, because the workflow installs dependencies from a yaml file instead of installing numpy manually. In MDAnalysis this flag works because of [these lines](https://github.com/MDAnalysis/mdanalysis/blob/8f9f402fe65d6a90e88b60b6da735c130425f00e/.github/workflows/gh-ci.yaml#L133):

```
conda_deps="${{ matrix.numpy }} ${MDA_CONDA_${{ matrix.run_type }}_DEPS}"
mamba install ${conda_deps}
```

which basically expand out to

```
conda_deps="numpy=1.16.0 scipy pytest ..."
mamba install numpy=1.16.0 scipy pytest ...
```

([mamba](https://github.com/mamba-org/mamba) is a drop-in replacement for conda).

I'm not sure if this is the most elegant solution (@IAlibay knows GH Actions much better than me), but you could have something like an optional numpy re-installation step:

```
- name: reinstall_numpy
- shell: bash
- run: |
  if [[ $numpy ]] ; then  # if the $numpy bash variable is set
  	conda -y uninstall numpy  # uninstall numpy without asking for confirmation
  	conda -y install $numpy  # install numpy version without asking for confirmation
  done
```

I can't remember if conda confirmation is necessary in GH actions.

lilyminium (changes requested):
Thanks for adding this, @ojeda-e. I've got some comments about numpy, specifically wondering if you a) need to import it in `setup.py`, and b) if you are testing the minimum version.

.github/workflows/CI.yaml
ojeda-e: Thanks. I checked with Python 3.7 and confirmed I got `numpy=1.16.0`. I didn't check for 3.6 because I took it for granted and now that I checked I saw `numpy=1.19.5`. For Python 3.8 and higher, I am getting `numpy=1.21.1` (reason why I also added the `1.20.2` version in the first submission of the PR.)
But then there is something I am still missing. If the workflow installs dependencies from the yaml file, as you said, why does MDAnalyis have >=1.17.3 in [the yaml file](https://github.com/MDAnalysis/mdanalysis/blob/develop/maintainer/conda/environment.yml), and [the CI](https://github.com/MDAnalysis/mdanalysis/blob/develop/.github/workflows/gh-ci.yaml) has the `min_numpy` with 1.16.0?

.github/workflows/CI.yaml
ojeda-e: Thanks, I missed this. Same below.

setup.py
IAlibay: I don't _think_ downstream packages need to ensure numpy import. Putting it in `install_requires` should be fine (assuming the minimum/maximum version differs from upstream MDAnalysis).

setup.py
```diff
@@ -12,12 +12,28 @@
...
     with open("README.md", "r") as handle:
         long_description = handle.read()
-except:
-    long_description = "\n".join(short_description[2:])
 
+    # set requirements
+    install_requires = [
+        'numpy>=1.16.0',
+        'mdanalysis>=2.0.0b0',
+        'mdanalysistests>=2.0.0',
```
IAlibay, line 35:
```suggestion
        'mdanalysistests>=2.0.0b0',
```
?

setup.py
ojeda-e: I will get rid of this, then. Thanks.

setup.py
```diff
@@ -12,12 +12,28 @@
 needs_pytest = {'pytest', 'test', 'ptr'}.intersection(sys.argv)
 pytest_runner = ['pytest-runner'] if needs_pytest else []
 
+# Check Python version
+if sys.version_info[:2] < (3, 6):
```
IAlibay, line 16: I'm not sure this is required either (given that you set python_requires).

ojeda-e:
Here is the summary of what I found (thanks @IAlibay for the guidance):
According to [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html#), a project should work with
- all minor versions of Python released in the 42 months prior to the project, and at minimum the last three minor versions.
- all minor versions of NumPy released in the 24 months prior to the project, and at minimum the last three minor versions.

According to this table from NEP29, the minimum NumPy version supported for MembraneCurvature should be  Python>=3.6 and Numpy>=1.15.


It also says:
"In setup.py, the python_requires variable should be set to the minimum supported version of Python. All supported minor versions of Python should be in the test matrix and have binary artifacts built for the release."

Now, following NEP29 would mean I have to add all the 12 minor versions of python which seems a bit excessive at this point. Also according to NEP29, I would have to use Numpy>=1.15, which makes me hesitate about what to do since MDAnalysis currently set a higher version minimum requirement.

I tried all the options that occurred to me to cover different versions of NumPy in the tests to check they work. So I failed and I'm giving up at this point.

In Python=3.6, the tests are running and passing with NumPy=1.19.5; In Python=3.7, tests are passing with Numpy=1.21.1 .So I am setting Numpy>=1.19.5 and Python>=3.6, and I will ignore NEP29 for now.
I still don't understand why if `install_requires` is enough, then the min version of Numpy is not the same across versions of Python.

Sorry for the annoying notifications @orbeckst @tylerjereddy @IAlibay @lilyminium , I cleaned the mess.

lilyminium:
I believe that that support table is the minimum versions that numpy will continue to support, instead of the release schedule. As we're now in July 2021, that would soon be Python 3.7+ and 1.18+. MDAnalysis tends to take a longer view of these things -- our last Python 2.7-supporting release was sometime this year, an entire year after the Python 2 end of life. There was some discussion about this on [Discord](https://discord.com/channels/807348386012987462/808088023957897258/849816892636921906) in June (probably easier to read on the [mailing list too](https://groups.google.com/d/msgid/mdnalysis-devel/46fca9d6-4b29-423a-a0c1-e925a123c4dfn%40googlegroups.com?utm_medium=email&utm_source=footer) -- if you're not subscribed already, it's a great way to track code-related discussions).

I think you are right that you can probably drop numpy 1.16 and 1.17 in following NEP29 (I believe 1.16 was required for MDAnalysis 1.0 for the 2.7 support). 1.19.5 is a bit far, though -- do your tests pass with lower versions? Usually having the most lax requirements possible is best, because that gives users more flexibility in building their Python environments.

> Now, following NEP29 would mean I have to add all the 12 minor versions of python which seems a bit excessive at this point. 
> I tried all the options that occurred to me to cover different versions of NumPy in the tests to check they work.

That's very thorough of you! Would you mind elaborating on what you tried and didn't work? These things are always [good to know for the future](https://www.livecomsjournal.org/post/130).

However, I'm not sure this much testing is necessary. I guess I think of the tests more as a way to make sure that I'm not using features that are too new, or old features that have been removed. So for the former I would just check that I can run my tests with my minimum numpy version, e.g. 1.16, and minimum Python version, e.g. 3.6. For the latter, I keep an eye out for deprecation warnings and eventually the tests will fail. For example, if you are testing Python 3.9, only numpy 1.20+ is supported; so if tests suddenly start failing here, then I know something might have changed in numpy 1.20. @MDAnalysis/gsoc-mentors please do chip in if you disagree or have more to add.

Also, in [semantic versioning](https://semver.org/), the versions go MAJOR.MINOR.PATCH. So for Python 3.7.1, the major version is 3, the minor version is 7, and the patch version is 1. Therefore when it says "All supported minor versions of Python should be in the test matrix", there are only four between [3.6, 3.7, 3.8, 3.9]. I think you're already testing all of these!

ojeda-e:
Thanks for the comments @lilyminium 
Yes, my bad I didn't elaborate on the issues.
My limitation is to test the lowest version of NumPy and different Python versions simulatenously. To clarify, individual tests pass, but what I couldn't manage to do is installing more than one specific version of NumPy for the tests in combination with the matrix of Python. I used different approaches, and in some of them _I think_  MDAnalysis rewrites the version of NumPy when it is installed, and when I go through non very orthodox approaches like the one here below, there is a side effect in the installation of MDAanalysis. For instance, something as you suggested (tried retrieving the comment but can't find it), slightly modified:
```
      shell: bash -l {0}
      run: |
        python -m pip install MDAnalysis==2.0.0b0
        python -m pip unistall -y numpy
        python -m pip install numpy=={{ NUMPY_VERSION }}
        python -m pip install . --no-deps
        conda list
```
it doesn't work. The error suggests MDAnalysis is not properly installed (I didn't reproduce this because it's ugly).

Now, if I try to use a matrix for the version of NumPy, to have Python>=3.6 and Numpy=1.16.0, as in the same style [as MDAnalysis](https://github.com/MDAnalysis/mdanalysis/blob/develop/.github/workflows/gh-ci.yaml), I got [this mamba error](https://github.com/MDAnalysis/membrane-curvature/runs/3114865674?check_suite_focus=true)
```
Looking for: ['numpy=1.16.0', 'pip', 'pytest', 'mmtf-python', 'biopython', 'networkx', 'cython', 'matplotlib-base', 'scipy', 'griddataformats', 'hypothesis', 'gsd', 'codecov', 'threadpoolctl']

Pinned packages:
  - python 3.8.5

Encountered problems while solving:
  - package numpy-1.16.0-py37_blas_openblash1522bff_1000 requires python >=3.7,<3.8.0a0, but none of the providers can be installed
```

- When I use numpy=1.17, it only fails for Python 3.6, and same error. (and for all windows :s )

With that error (And excluding the fact that windows is windows and it fails), I said, oh ok, so maybe I do have a min version of Python. So first, changed to 3.7 min, even if that would contradict tests from MDAnalysis (because they pass with 3.6). So,

- When I use numpy=1.17 and Python 3.7  minimum, I [get a similar error](https://github.com/MDAnalysis/membrane-curvature/runs/3115022000?check_suite_focus=true). (all windows fails)
```
Looking for: ['numpy=1.16.0', 'pip', 'pytest', 'mmtf-python', 'biopython', 'networkx', 'cython', 'matplotlib-base', 'scipy', 'griddataformats', 'hypothesis', 'gsd', 'codecov', 'threadpoolctl']

Pinned packages:
  - python 3.8.5

Encountered problems while solving:
  - package numpy-1.16.0-py37_blas_openblash1522bff_1000 requires python >=3.7,<3.8.0a0, but none of the providers can be installed
```
So, in the end, it doesn't mean Python=3.6 doesn't work, it's more like I am doing something wrong that I can't see. Maybe @IAlibay has an idea of how to troubleshoot this and make Python=3.6 and numpy=1.16.0 work simulatenously with other versions of Python?

So this is the summary of how I gave up and ended up with Numpy=1.19.5. Also, Thanks for this 
> For example, if you are testing Python 3.9, only numpy 1.20+ is supported; so if tests suddenly start failing here, then I know something might have changed in numpy 1.20.

.github/workflows/CI.yaml
```diff
@@ -60,8 +72,16 @@ jobs:
         conda list
 
 
-    - name: Run tests
+    - name: Install numpy
```
IAlibay, line 75: The way this workflow is laid out, installing numpy here isn't really going to work (or if anything it'll end up problematic).

You have to consider these runner workflows to be serial in time. So here you're doing the following:

1. Checking out the repo
2. Getting some info about the system (`df -h`, etc...)
3. Creating an conda environment backed on your test_env.yaml (which will auto-pull numpy, etc.. because you've either defined them or they get pulled by MDA)
4. Installed your package
5. Installed a new version of numpy
6. Ran the tests

If you want to test pinning to an older version of numpy, at least for now, my suggestion is just to create a separate environment yaml file that pins to the lowest package versions you want to support.

Then add a `env-file` entry to your runner matrix where for the normal tests it points to your normal `test_env.yaml`, and for the minimum version it points to say `test_min_env.yaml`. Then use that variable to feed the `environment-file` entry on the `conda-incubator` step.

tylerjereddy:
There's also something to be said for using `pyproject.toml` these days instead of leaning too heavily on `setup.py` for some things, but I can't remember if even core MDA has started down that path just yet.

i.e., https://setuptools.readthedocs.io/en/latest/build_meta.html

That said, you could spend all your time trying to get packaging right, bit of a pain in Python still..

IAlibay:
> There's also something to be said for using `pyproject.toml` these days instead of leaning too heavily on `setup.py` for some things, but I can't remember if even core MDA has started down that path just yet.

Not yet (as far as I know). @tylerjereddy at some point I'd like to pick your brains about all the potential changes to core MDA's setup that you would recommend. I've been thinking it might be a good idea to see if we can get an NF small development grant to do an overhaul of setup + automated weekly/monthly dev builds.
 
> That said, you could spend all your time trying to get packaging right, bit of a pain in Python still..

In my opinion here, we can probably defer this to a new issue / future work, given that:

a) *for now* we are following the CMS cookiecutter (which hasn't adopted things like `pyproject.toml`)
b) we aren't overly concerned about releases yet (but rather just making sure that CI works).

.github/workflows/CI.yaml
```diff
@@ -39,17 +41,54 @@ jobs:
...
+        python-version: ${{ matrix.python-version-min }}
+        environment-file: devtools/conda-envs/test_min.yaml
+        channels: conda-forge, defaults
+
+        activate-environment: test
+        auto-update-conda: false
+        auto-activate-base: false
+        show-channel-urls: true
+        mamba-version: "*"   
+
```
ojeda-e, line 56: I think this is the source of error. I forgot to delete mamba.

PR #51, opened 2021-07-21
Possible CI solution [PR #49 target]

IAlibay:
## Description
As discussed here is my solution (assuming CI doesn't complain) to the CI issues in PR #49 

## To do
- [ ] Remove extra branch in PR (currently used to test)

## Status
- [ ] Ready to go

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -31,6 +31,7 @@ jobs:
         env_file: [env, ]
         include:
           - name: minimum_requirements
+            os: ubuntu-latest
             python-version: 3.6
             env_file: min
 
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -21,8 +21,8 @@ concurrency:
 
 jobs:
   test:
-    name: Test on ${{ matrix.os }}, Python ${{ matrix.python-version }}
-    runs-on: ubuntu-latest
+    name: "Test on ${{ matrix.os }}, Python ${{ matrix.python-version }}"
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -7,6 +7,8 @@ on:
   pull_request:
     branches:
       - "main"
+      # remove before merging
+      - "issue49"
   # schedule:
     # Nightly tests run on master by default:
     #   Scheduled workflows run on the latest commit on the default or base branch.
@@ -25,12 +27,15 @@ jobs:
       fail-fast: false
       matrix:
         os: [macOS-latest, ubuntu-latest, windows-latest]
-        python-version-min: [3.6, 3.7]
         python-version: [3.7, 3.8, 3.9]
-        
+        env_file: [env, ]
+        include:
+          - name: minimum_requirements
+            python-version: 3.6
+            env_file: min
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
 
     - name: Additional info about the build
       shell: bash
@@ -39,23 +44,19 @@ jobs:
         df -h
         ulimit -a
 
-
     # More info on options: https://github.com/conda-incubator/setup-miniconda
     # An environment for the minimum versions 
     - uses: conda-incubator/setup-miniconda@v2
       with:
-        python-version: ${{ matrix.python-version-min }}
-        environment-file: devtools/conda-envs/test_min.yaml
+        python-version: ${{ matrix.python-version }}
+        environment-file: devtools/conda-envs/test_${{ matrix.env_file }}.yaml
         channels: conda-forge, defaults
-
         activate-environment: test
         auto-update-conda: false
         auto-activate-base: false
         show-channel-urls: true
-        mamba-version: "*"   
     
     - name: Install package
-
       # conda setup requires this special shell
       shell: bash -l {0}
       run: |
@@ -65,7 +66,6 @@ jobs:
     - name: Run tests
       # conda setup requires this special shell
       shell: bash -l {0}
-
       run: |
         pytest -v --cov=membrane_curvature --cov-report=xml --color=yes membrane_curvature/tests/
 
@@ -75,39 +75,5 @@ jobs:
         file: ./coverage.xml
         flags: unittests
         name: codecov-${{ matrix.os }}-py${{ matrix.python-version }}
+        fail_ci_if_error: True
 
-
-    # Another environment for the the highest versions
-    - uses: conda-incubator/setup-miniconda@v2
-      with:
-        python-version: ${{ matrix.python-version }}
-        environment-file: devtools/conda-envs/test_env.yaml
-        channels: conda-forge, defaults
-
-        activate-environment: test
-        auto-update-conda: false
-        auto-activate-base: false
-        show-channel-urls: true
-        mamba-version: "*"
-
-    - name: Install package
-
-      # conda setup requires this special shell
-      shell: bash -l {0}
-      run: |
-        python -m pip install . --no-deps
-        conda list
-
-    - name: Run tests
-      # conda setup requires this special shell
-      shell: bash -l {0}
-
-      run: |
-        pytest -v --cov=membrane_curvature --cov-report=xml --color=yes membrane_curvature/tests/
-
-    - name: CodeCov
-      uses: codecov/codecov-action@v1
-      with:
-        file: ./coverage.xml
-        flags: unittests
-        name: codecov-${{ matrix.os }}-py${{ matrix.python-version }}
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -22,7 +22,7 @@ concurrency:
 jobs:
   test:
     name: Test on ${{ matrix.os }}, Python ${{ matrix.python-version }}
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -7,7 +7,6 @@ on:
   pull_request:
     branches:
       - "main"
-      # remove before merging
       - "issue46"
   # schedule:
     # Nightly tests run on master by default:
@@ -18,6 +17,7 @@ on:
 concurrency:
   group: "${{ github.ref }}-${{ github.head_ref }}"
   cancel-in-progress: true
+
 
 jobs:
   test:
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -8,7 +8,7 @@ on:
     branches:
       - "main"
       # remove before merging
-      - "issue49"
+      - "issue46"
   # schedule:
     # Nightly tests run on master by default:
     #   Scheduled workflows run on the latest commit on the default or base branch.
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -7,7 +7,6 @@ on:
   pull_request:
     branches:
       - "main"
-      - "issue46"
   # schedule:
     # Nightly tests run on master by default:
     #   Scheduled workflows run on the latest commit on the default or base branch.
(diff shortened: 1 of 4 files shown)


IAlibay:
Ah yes, the ever so descriptive gh actions error messages: "Error when evaluating 'runs-on' for job 'test'. (Line: 25, Col: 14): Unexpected value ''

Sometimes I wonder if it's possible worse than awk's interpretor....

Apologies for the many commits as I work out what character I'm missing...

IAlibay:
hand-defining ubuntu works, now to see what's up with matrix.os...

.github/workflows/CI.yaml
```diff
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - "main"
+      - "issue46"
```
IAlibay, line 10: to be removed once tests pass

.github/workflows/CI.yaml
```diff
@@ -25,12 +27,15 @@ jobs:
       fail-fast: false
       matrix:
         os: [macOS-latest, ubuntu-latest, windows-latest]
-        python-version-min: [3.6, 3.7]
         python-version: [3.7, 3.8, 3.9]
-        
+        env_file: [env, ]
```
IAlibay, line 31: Here we add this extra variable to pick up the right environment file: `env` for the normal `test_env.yaml` and `min` for `test_min.yaml`.

In this case all the normal jobs in the matrix (which run all 3 OS versions + all the python versions) will use `test_env.yaml`.

.github/workflows/CI.yaml
```diff
@@ -25,12 +27,15 @@ jobs:
...
+        include:
```
IAlibay, line 32: Using `include` we can add extra jobs that aren't part of the main matrix tests. Here we add an extra job that runs python 3.6 on ubuntu and uses the minimum environment yaml file.

.github/workflows/CI.yaml
```diff
@@ -25,12 +27,15 @@ jobs:
...
+          - name: minimum_requirements
+            python-version: 3.6
+            env_file: min
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
```
IAlibay, line 39: bumped up version because why not

.github/workflows/CI.yaml
```diff
@@ -39,23 +44,19 @@ jobs:
...
 
-
     # More info on options: https://github.com/conda-incubator/setup-miniconda
     # An environment for the minimum versions 
     - uses: conda-incubator/setup-miniconda@v2
       with:
-        python-version: ${{ matrix.python-version-min }}
-        environment-file: devtools/conda-envs/test_min.yaml
+        python-version: ${{ matrix.python-version }}
+        environment-file: devtools/conda-envs/test_${{ matrix.env_file }}.yaml
```
IAlibay, line 53: we use the `env_file` variable to selectively choose which environment file is passed through.

.github/workflows/CI.yaml
```diff
@@ -75,39 +75,5 @@ jobs:
         file: ./coverage.xml
         flags: unittests
         name: codecov-${{ matrix.os }}-py${{ matrix.python-version }}
+        fail_ci_if_error: True
```
IAlibay, line 79: Added this to have CI fail when there's an issue with codecov

.github/workflows/CI.yaml
```diff
@@ -75,39 +75,5 @@ jobs:
...
 
-
-    # Another environment for the the highest versions
```
IAlibay, line 80: Each job runs though the steps once, you don't need to do multiple set of steps, instead we have the extra job added to `include` that will run minimum tests.

devtools/conda-envs/test_env.yaml
```diff
@@ -5,7 +5,7 @@ channels:
 dependencies:
   - python
   - pip
-  - numpy==1.18.0
+  - numpy
```
IAlibay, line 8: no need to pin to a version here, we can assume that the highest version will be picked up (with the minimum tests you then have reasonable confidence that the whole range of numpy versions will work).

devtools/conda-envs/test_min.yaml
```diff
@@ -1,17 +1,17 @@
-name: test_min
+name: test
```
IAlibay, line 1: Renamed to `test` since we explicitly tell `conda-incubator/setup-miniconda@v2` to active an environment named `test`

devtools/conda-envs/test_min.yaml
```diff
@@ -1,17 +1,17 @@
...
 channels:
   - conda-forge
   - defaults
 dependencies:
   - python
   - pip
-  - sphinx
   - numpy==1.16.0
+  - sphinx
```
IAlibay, line 9: this was just a switch around to make it match better with the order in `test_env.yaml`

setup.py
```diff
@@ -12,17 +12,12 @@
 needs_pytest = {'pytest', 'test', 'ptr'}.intersection(sys.argv)
 pytest_runner = ['pytest-runner'] if needs_pytest else []
 
-if __name__ == "__main__":
-
+try:
```
IAlibay, line 15: no need to use __name__ == "__main__" as far as I'm aware. We use it MDA core because of all the extra bits we do in the setup file, but I think here it's ok to just stick to what the CMS cookiecutter gave as the first solution.

setup.py
```diff
@@ -47,11 +42,19 @@
     include_package_data=True,
 
     # Allows `setup.py test` to work correctly with pytest
-    setup_requires=['numpy>=1.19.5'] + pytest_runner,
+    setup_requires=[] + pytest_runner,
```
IAlibay, line 45: numpy isn't required for setup (I believe we need it in MDAnalysis core because we call numpy as part of our setup procedure)

setup.py
```diff
@@ -47,11 +42,19 @@
...
 
     # Additional entries you may want simply uncomment the lines you want and fill in the data
     # url='http://www.my_package.com',  # Website
-    # install_requires=[],              # Required packages, pulls from pip if needed; do not use for Conda deployment
+    install_requires = [
```
IAlibay, line 49: move the install_requires bits here (I don't have an explanation, I _think_ that's just how it works)

setup.py
```diff
@@ -47,11 +42,19 @@
...
+        'numpy>=1.16.0',
+        'mdanalysis>=2.0.0b0',
+    ],
+
+    # packages required to run tests
+    tests_require = [
```
IAlibay, line 55: You can add a `tests_require` entry to force it to make sure that `pytest` is available if you want to run the tests (not super necessary, but it's nice enough to have).

setup.py
```diff
@@ -12,17 +12,12 @@
...
     with open("README.md", "r") as handle:
         long_description = handle.read()
+except:
+    long_description = "\n".join(short_description[2:])
 
-    # set requirements
-    install_requires = [
-        'numpy>=1.19.5',
-        'mdanalysis>=2.0.0b0',
-        'mdanalysistests>=2.0.0b0',
```
IAlibay, line 24: MDAnalysisTests aren't used anywhere so I just removed this from `install_requires`

IAlibay (commented):
@ojeda-e here's my explanation for all the bits I changed, let me know if I can clarify things.

devtools/conda-envs/test_env.yaml
ojeda-e: Ok, I think my understanding of this was quite the opposite. I am missing a key point here. The premise I had was "I am going to test the minimum version of Numpy for each version of Python. Then I can identify what's my bottom end, and therefore, I identify what is the min version supported in my repo". 
You say 
> we can assume that the highest version will be picked up

Why would I be interested in the highest version? Shouldn't I be targeting the min version for each version of Python?

setup.py
ojeda-e: Ok, I was going to leave it to use it in the long term, but not currently using it is a good point.

setup.py
ojeda-e: This is another thing I completely misunderstood. Maybe you could try to explain what is your criteria to determine what is required in the setup? If my repo depends fully on NumPy... how come it is not included in the setup_requires? 🤯  Does that mean that Python makes a difference between "What I need to install this repo" vs " what I need to install to make the code in the repo work"?

.github/workflows/CI.yaml
ojeda-e: According to the [Numpy release notes](https://numpy.org/devdocs/release/1.16.0-notes.html), the 1.16.0 version of python is supported in Python 3.5-3.7.
Would you please help me to understand why are we insisting on python=3.6 and Numpy=1.16.0 only? Shouldn't it be tested for 
- Python = 3.6 *And* 3.7 with Numpy=1.16.0?
Similarly, shouldn't we test 
- Python = 3.8  with Numpy=1.18.0,
and
-Python = 3.9 with Numpy=1.19.0?

IAlibay: Sorry this is going to be a bit messy to explain :(

So the way I see is that numpy and python are two variables that you can treat separately.

1. It's ok (at least for the sake of sanity) to make the assumption that for any given numpy minor version behaves the exact same way across the multiple python minor versions it supported at the time of release (that's at the very least the assumption that libraries like numpy/scipy and even MDA make). With this in mind, you can just test one version and say it works. 
2. For 1.16.0, here I set it to 3.6 because py3.6 + numpy 1.16.0 will pick up a ton of smaller packages (which the bigger packages we define rely on), that might be lower in version than py3.7 + numpy 1.16.0. In practice this has a near zero impact, but it doesn't hurt.
3. Technically you could test all the numpy versions, but it's very expensive and with libraries like numpy you can kinda make the assumption that as long as the difference between the lowest possible version you can support (1.16.0) and the newest version (1.21.0) isn't too large, then the API has progressively changed between the min & max version. That is to say that anything in 1.21.0 that isn't in 1.16.0 will have been introduced in all the intermediate minor versions. Based on that, you can make the assumption that anything that if things work with the minimum and maximum versions, then everything in between will work fine too.

devtools/conda-envs/test_env.yaml
IAlibay: My response https://github.com/MDAnalysis/membrane-curvature/pull/51#discussion_r674408301 above probably answers this to an extent.

The short answer is that as long as you test the two extremes you can be somewhat confident that the intermediate versions will behave properly, especially if you eventually follow NEP29 and don't have too large a gap between numpy versions.

setup.py
IAlibay: So setuptools (which is the thing that's used to run these setup files) is a lot of magic that I haven't had the chance to full decipher, but yes it (tries to) discriminates between what's required to "setup the repo" and what's required to "run the code".

Essentially here is the entry for setup_requires: "A string or list of strings specifying what other distributions need to be present in order for the setup script to run. setuptools will attempt to obtain these (even going so far as to download them using EasyInstall) before processing the rest of the setup script or commands. "

My understanding is that most time you really don't need anything to run setup. In the MDA core library we need numpy because we compile some Cython code that requires numpy's header files during the setup process.

devtools/conda-envs/test_env.yaml
ojeda-e: Thanks, I was about to reply along the same lines. So, to finish this topic, would you please confirm the following statement is correct? "The min version I am going to put on my setup.py is the min version I test, and I will assume that all in between that lowest end and the highest will also work." (I get the highest version from the `test_env.yaml` when I create the environment)

setup.py
IAlibay: Assuming you only need them for the tests I guess you can it in tests_require eventually? MDAnalysisTests is actually quite large (~ 40 MB), so if you can avoid users downloading it they'll appreciate it.

devtools/conda-envs/test_env.yaml
IAlibay: Yeah, without pinning `test_env.yaml` should pick up the latest version assuming it's compatible with all your dependencies.

Technically if you wanted to be super accurate / strict you could also put a maximum version in your setup.py, but just putting the lowest version you've tested on is fine as long as you've tested the most recent version at the time of release.

IAlibay:
@ojeda-e please do review & squash-merge when you want. After that #46 will probably be ok to merge.

ojeda-e:
Thanks for your help @IAlibay. I definitely had missing pieces to solve this puzzle.

PR #52, opened 2021-07-22
Surface Interpolation Function [Fixes Issue45]

### LINKED ISSUE #45: Undefined values in bins occupied by insertion in membrane.
When calculating `get_z_surface` undefined values in bins inside the embedded element (i.e. protein) may arise since no `z` coordinates populate bins in the grid. Such undefined values spread in the array during the calculation of curvature.

One possible solution is to interpolate the bins occupied by the protein. I would avoid using `pandas` although that approach seems the most straightforward. Check `np.interpolate`.

ojeda-e:
This is a starting point to solve #45 

Changes in this PR:

- [x] Added function to calculate interpolation of z_surface (`surface_interpolation`). Includes docstrings.
- [x] Added tests of `surface_interpolation`. 

As suggested by @orbeckst in this [comment](https://github.com/MDAnalysis/membrane-curvature/issues/45#issuecomment-881972548), the function here included uses NumPy. Thanks for encouraging numpy! 

Surfaces included in tests are dummy_arrays with the following characteristics:
- size 3x3 with all values 150 and one `np.nan`.
- size 4x3 with all values 150 and two `np.nan`.
- size 4x4 with all values 150 and two `np.nan`.
- size 3x3 array 3x3  with lots of `np.nan`. 💪🏽 

Probably I'll be asked to add more tests, but this is ok to start. Probably @lilyminium will give me lots of ideas. 😅 

 Thanks.

### DIFF
--- a/membrane_curvature/base.py
+++ b/membrane_curvature/base.py
@@ -12,7 +12,7 @@ surfaces derived from a selection of reference.
 
 import numpy as np
 import warnings
-from .surface import get_z_surface
+from .surface import get_z_surface, surface_interpolation
 from .curvature import mean_curvature, gaussian_curvature
 
 import MDAnalysis
@@ -32,7 +32,7 @@ class MembraneCurvature(AnalysisBase):
     ----------
     universe : Universe or AtomGroup
         An MDAnalysis Universe object.
-    select : str or iterable of str, optional. 
+    select : str or iterable of str, optional.
         The selection string of an atom selection to use as a
         reference to derive a surface.
     wrap : bool, optional
@@ -57,13 +57,13 @@ class MembraneCurvature(AnalysisBase):
     results.gaussian_curvature : ndarray
         Gaussian curvature associated to the surface.
         Arrays of shape (`n_frames`, `n_x_bins`, `n_y_bins`)
-    results.average_z_surface : ndarray 
-        Average of the array elements in `z_surface`. 
+    results.average_z_surface : ndarray
+        Average of the array elements in `z_surface`.
         Each array has shape (`n_x_bins`, `n_y_bins`)
-    results.average_mean_curvature : ndarray 
+    results.average_mean_curvature : ndarray
         Average of the array elements in `mean_curvature`.
         Each array has shape (`n_x_bins`, `n_y_bins`)
-    results.average_gaussian_curvature: ndarray 
+    results.average_gaussian_curvature: ndarray
         Average of the array elements in `gaussian_curvature`.
         Each array has shape (`n_x_bins`, `n_y_bins`)
 
@@ -120,7 +120,8 @@ class MembraneCurvature(AnalysisBase):
                  n_x_bins=100, n_y_bins=100,
                  x_range=None,
                  y_range=None,
-                 wrap=True, **kwargs):
+                 wrap=True,
+                 interpolation=False, **kwargs):
 
         super().__init__(universe.universe.trajectory, **kwargs)
         self.ag = universe.select_atoms(select)
@@ -129,6 +130,7 @@ class MembraneCurvature(AnalysisBase):
         self.n_y_bins = n_y_bins
         self.x_range = x_range if x_range else (0, universe.dimensions[0])
         self.y_range = y_range if y_range else (0, universe.dimensions[1])
+        self.interpolation = interpolation
 
         # Raise if selection doesn't exist
         if len(self.ag) == 0:
@@ -156,7 +158,6 @@ class MembraneCurvature(AnalysisBase):
             logger.warn(msg)
 
     def _prepare(self):
-        # Initialize empty np.array with results
         self.results.z_surface = np.full((self.n_frames,
                                           self.n_x_bins,
                                           self.n_y_bins), np.nan)
@@ -166,6 +167,9 @@ class MembraneCurvature(AnalysisBase):
         self.results.gaussian = np.full((self.n_frames,
                                          self.n_x_bins,
                                          self.n_y_bins), np.nan)
+        self.results.interpolated_z_surface = np.full((self.n_frames,
+                                                       self.n_x_bins,
+                                                       self.n_y_bins), np.nan)
 
     def _single_frame(self):
         # Apply wrapping coordinates
@@ -179,8 +183,20 @@ class MembraneCurvature(AnalysisBase):
                                                                   y_range=self.y_range)
         self.results.mean[self._frame_index] = mean_curvature(self.results.z_surface[self._frame_index])
         self.results.gaussian[self._frame_index] = gaussian_curvature(self.results.z_surface[self._frame_index])
+        # Perform interpolation fo surface
+        if self.interpolation:
+            # Populate a slice with np.arrays of interpolated surface
+            self.results.interpolated_z_surface[self._frame_index] = surface_interpolation(
+                self.results.z_surface[self._frame_index])
 
     def _conclude(self):
         self.results.average_z_surface = np.nanmean(self.results.z_surface, axis=0)
         self.results.average_mean = np.nanmean(self.results.mean, axis=0)
         self.results.average_gaussian = np.nanmean(self.results.gaussian, axis=0)
+        if self.interpolation:
+            self.results.average_interpolated_z_surface = np.nanmean(
+                self.results.interpolated_z_surface[self._frame_index], axis=0)
+            self.results.average_interpolated_mean = np.nanmean(mean_curvature(
+                self.results.interpolated_z_surface[self._frame_index]), axis=0)
+            self.results.average_interpolated_gaussian = np.nanmean(gaussian_curvature(
+                self.results.interpolated_z_surface[self._frame_index]), axis=0)
--- a/membrane_curvature/surface.py
+++ b/membrane_curvature/surface.py
@@ -131,3 +131,60 @@ def normalized_grid(grid_z_coordinates, grid_norm_unit):
     z_normalized = grid_z_coordinates / grid_norm_unit
 
     return z_normalized
+
+
+def interpolation_by_array(array_surface):
+    """
+        Interpolates values contained in array_surface over axi
+
+    Parameters
+    ----------
+
+    array_surface: numpy.ndarray
+        Numpy array of floats of shape (`n_x_bins`, `n_y_bins`)
+
+    Returns
+    -------
+    interpolated_array: np.ndarray
+        Returns interpolated array.
+        Numpy array of shape (`n_x_bins` x `n_y_bins`,)
+
+    """
+
+    # create mask for nans
+    mask_nans = np.isnan(array_surface)
+
+    # index of array_surface
+    index_array = np.arange(array_surface.shape[0])
+
+    # interpolate values in array
+    interpolated_array = np.interp(index_array,
+                                   np.flatnonzero(~mask_nans),
+                                   array_surface[~mask_nans])
+
+    return interpolated_array
+
+
+def surface_interpolation(array_surface):
+    """
+        Calculates interpolation 
+
+    Parameters
+    ----------
+
+    array_surface: np.ndarray
+        Numpy array of floats of shape (`n_x_bins`, `n_y_bins`)
+
+    Returns
+    -------
+
+    Returns interpolated surface
+        Interpolated surface derived from set of coordinates 
+        in grid of `x_range, y_range` dimensions.
+        Numpy array of floats of shape (`n_x_bins`, `n_y_bins`)
+
+    """
+
+    interpolated_surface = np.apply_along_axis(interpolation_by_array, axis=0, arr=array_surface)
+
+    return interpolated_surface
--- a/membrane_curvature/surface.py
+++ b/membrane_curvature/surface.py
@@ -55,7 +55,7 @@ def get_z_surface(coordinates, n_x_bins=10, n_y_bins=10, x_range=(0, 100), y_ran
 
     Parameters
     ----------
-    coordinates : np.ndarray 
+    coordinates : np.ndarray
         Coordinates of AtomGroup. NumPy array of shape=(n_atoms, 3).
     n_x_bins : int.
         Number of bins in grid in the `x` dimension. 
--- a/membrane_curvature/surface.py
+++ b/membrane_curvature/surface.py
@@ -39,7 +39,7 @@ def derive_surface(atoms, n_cells_x, n_cells_y, max_width_x, max_width_y):
 
     Returns
     -------
-    z_coordinates: numpy.ndarray
+    z_coordinates: np.ndarray
         Average z-coordinate values. Return Numpy array of floats of
         shape `(n_cells_x, n_cells_y)`.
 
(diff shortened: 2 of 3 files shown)


membrane_curvature/tests/test_membrane_curvature.py
```diff
@@ -208,6 +209,39 @@ def test_get_z_surface(x_bin, y_bin, x_range, y_range, expected_surface):
...
+    # array 3x3 with all 150 and one nan
+    (np.array(([150., 150., 150.],
+               [150., np.nan, 150.],
+               [150., 150., 150.])),
+     np.full((3, 3), 150.)),
+    # array 4x3 with all 150 and two nans
+    (np.array([[150.,  150, 150.,  150.],
+               [150., np.nan, np.nan,  150.],
+               [150., 150., 150.,  150.]]),
+     np.array([[150., 150,  150.,  150.],
```
ojeda-e, line 222: I just noticed I didn't use `np.full` here. Booo.

membrane_curvature/surface.py
```diff
@@ -97,11 +97,68 @@ def normalized_grid(grid_z_coordinates, grid_norm_unit):
...
+
+def surface_interpolation(array_surface):
+    """
+        Calculates interpolation
+
+    Parameters
+    ----------
+
+    array_surface: np.ndarray
+        Numpy array of floats of shape (`n_x_bins`, `n_y_bins`)
```
tylerjereddy, line 150: Here and in a few other places: Numpy -> NumPy

Could also just omit the NumPy specification or use `np.array` instead perhaps.

membrane_curvature/surface.py
```diff
@@ -97,11 +97,68 @@ def normalized_grid(grid_z_coordinates, grid_norm_unit):
...
+    # create mask for nans
+    mask_nans = np.isnan(array_surface)
+
+    # index of array_surface
+    index_array = np.arange(array_surface.shape[0])
+
+    # interpolate values in array
+    interpolated_array = np.interp(index_array,
+                                   np.flatnonzero(~mask_nans),
+                                   array_surface[~mask_nans])
```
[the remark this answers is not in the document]
ojeda-e: I completely missed it and it's a good point. You mean, having this instead, correct?

```
   mask_nans = ~np.isnan(array_surface) #invert original

   index_array = np.arange(array_surface.shape[0])

   interpolated_array = np.interp(index_array,
                                  np.flatnonzero(mask_nans),   #not inverted
                                  array_surface[mask_nans]).   #not inverted
```

membrane_curvature/tests/test_membrane_curvature.py
```diff
@@ -208,6 +209,39 @@ def test_get_z_surface(x_bin, y_bin, x_range, y_range, expected_surface):
...
+               [150., 150., 150.,  150.]])),
+    # array 3x3  with lots of nans
+    (np.array([[np.nan, np.nan, 150.],
+              [150, np.nan, 150.],
+              [np.nan, 150., np.nan]]),
+     np.full((3, 3), 150.))
+])
+def test_surface_interpolation(dummy_surface, expected_interpolated_surface):
+    surface = surface_interpolation(dummy_surface)
+    assert_almost_equal(surface, expected_interpolated_surface)
```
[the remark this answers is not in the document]
ojeda-e: Thanks. I remember @orbeckst [mentioned](https://github.com/MDAnalysis/membrane-curvature/pull/13#discussion_r651379750) that  MDAnalysis typically uses `assert_almost_equal`. Maybe we can have his opinion on what would be best here?

membrane_curvature/surface.py
```diff
@@ -97,11 +97,68 @@ def normalized_grid(grid_z_coordinates, grid_norm_unit):
...
+    -------
+
+    Returns interpolated surface
+        Interpolated surface derived from set of coordinates
+        in grid of `x_range, y_range` dimensions.
+        Numpy array of floats of shape (`n_x_bins`, `n_y_bins`)
+
+    """
+
+    interpolated_surface = np.apply_along_axis(interpolation_by_array, axis=0, arr=array_surface)
```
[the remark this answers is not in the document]
ojeda-e: We haven't really assessed performance. I'll skip it for the moment, but will keep it in mind.

membrane_curvature/tests/test_membrane_curvature.py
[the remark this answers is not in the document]
orbeckst: MDAnalysis uses `assert_almost_equal` in preference to `assert_array_almost_equal`. `assert_allclose` might be more recent that when we started using numpy tests so MDAnalysis might just be outdated and should be switching eventually. Follow @tylerjereddy 's advice :-).

membrane_curvature/tests/test_membrane_curvature.py
```diff
@@ -132,19 +133,19 @@ def small_grofile():
...
     for k, k_test in zip(MEMBRANE_CURVATURE_DATA['gaussian_curvature']['small'], K_test):
-        assert_almost_equal(k, k_test)
+        assert_allclose(k, k_test)
 
 
 def test_mean_curvature_small():
     H_test = mean_curvature(MEMBRANE_CURVATURE_DATA['z_ref']['small'])
     for h, h_test in zip(MEMBRANE_CURVATURE_DATA['mean_curvature']['small'], H_test):
-        assert_almost_equal(h, h_test)
+        assert_allclose(h, h_test, rtol=6)
```
ojeda-e, line 142: had to set `rtol=6` here and in the next test otherwise, the tests fail.

ojeda-e:
The pushed changes include:
- [x] replaced `assert_almost_equal` by `assert_allclose`. For [two assertions](https://github.com/MDAnalysis/membrane-curvature/pull/52/files#r683844953) I set tolerance to `rtol=6` otherwise the tests fail. Not sure you are ok with this change. (Thanks @tylerjereddy for the suggestion and @orbeckst for confirming)
- [x] added interpolation to `base.py` and tests for grids of different sizes and with a different number of undefined values. 
- [x] In `base.py`, kept the data for `z_surface` and `interpolated_z_surface` as suggested by @lilyminium.

I know `base.py` could be written better, but maybe we can leave that improvement for later and work on the essentials for now if that's ok. 
Hopefully I am not missing relevant things here. Thanks!

membrane_curvature/tests/test_membrane_curvature.py
lilyminium: Oh dear, that's a very large `rtol`. The default value is 1e-8, for reference. Are you sure you want tests to pass if any calculated value is within 600% of the target value?

lilyminium: If it used to pass with `assert_almost_equal` (where the `decimal=7` by default), you could try setting `atol=1e-7` in `allclose`?

ojeda-e: The test passes with `assert_allclose(h, h_test)` (without specifying `rtol=6`).

membrane_curvature/tests/test_membrane_curvature.py
```diff
@@ -132,19 +133,19 @@ def small_grofile():
...
 
 
 def test_gaussian_curvature_all():
     K_test = gaussian_curvature(MEMBRANE_CURVATURE_DATA['z_ref']['all'])
     for k, k_test in zip(MEMBRANE_CURVATURE_DATA['gaussian_curvature']['all'], K_test):
-        assert_almost_equal(k, k_test)
+        assert_allclose(k, k_test, rtol=6)
```
ojeda-e, line 148: It should be `assert_allclose(k, k_test, atol=1e-7)`

ojeda-e:
@lilyminium sorry, it's my fault because I made a bit of a mess here.
To clarify. 
`def test_mean_curvature_small` passes with `assert_allclose`.  I don't know why I ended up adding `rtol` it should go.  (I added the comment [here](https://github.com/MDAnalysis/membrane-curvature/pull/52/files#r684656934))

The problem is in `def test_gaussian_curvature_all`. The test passes with `rtol=4` or `atol=1e-7`. From the [docs](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose) it says `allclose` "... compares the difference between actual and desired to atol + rtol * abs(desired)."
So probably better have `assert_allclose(k, k_test, atol=1e-7)` ?

lilyminium:
> So probably better have assert_allclose(k, k_test, atol=1e-7) ?

I think so. Even if you know yourself that the 400% difference is small because the real value is small, it communicates to future users that you expect the degree of difference to be around 0.0000001.

membrane_curvature/base.py
```diff
@@ -179,8 +183,20 @@ def _single_frame(self):
...
+            self.results.interpolated_z_surface[self._frame_index] = surface_interpolation(
+                self.results.z_surface[self._frame_index])
 
     def _conclude(self):
         self.results.average_z_surface = np.nanmean(self.results.z_surface, axis=0)
         self.results.average_mean = np.nanmean(self.results.mean, axis=0)
         self.results.average_gaussian = np.nanmean(self.results.gaussian, axis=0)
+        if self.interpolation:
+            self.results.average_interpolated_z_surface = np.nanmean(
+                self.results.interpolated_z_surface[self._frame_index], axis=0)
```
lilyminium, line 198:
```suggestion
                self.results.interpolated_z_surface, axis=0)
```

And with the others too -- if you specify `frame_index`, you'll wind up only getting the mean of the last frame.

membrane_curvature/surface.py
```diff
@@ -131,3 +131,60 @@ def normalized_grid(grid_z_coordinates, grid_norm_unit):
     z_normalized = grid_z_coordinates / grid_norm_unit
 
     return z_normalized
+
+
+def interpolation_by_array(array_surface):
+    """
+        Interpolates values contained in `array_surface` over axis
```
lilyminium, line 138: Over which axis?

membrane_curvature/surface.py
```diff
@@ -131,3 +131,60 @@ def normalized_grid(grid_z_coordinates, grid_norm_unit):
...
+
+    # index of array_surface
+    index_array = np.arange(array_surface.shape[0])
+
+    # interpolate values in array
+    interpolated_array = np.interp(index_array,
+                                   np.flatnonzero(mask_nans),
+                                   array_surface[mask_nans])
+
+    return interpolated_array
```
lilyminium, line 165: I'm not quite understanding this. `mask_nans` is 2D, with shape (n_x, n_y). `index_array` is 1D, with shape `n_x`. `np.flatnonzero(mask_nans)` and `array_surface[mask_nans]` will be 1D, of length between 0 to n_x * n_y. So I *think* that means that you are only ever operating on the first row of `array_surface`?

I guess my question is then, why is `array_surface` 2D? Why not just pass in one row at a time? In addition, as it is a 2D surface, why use a 1D interpolation function? Why not a [2D one](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp2d.html)?

membrane_curvature/surface.py
```diff
@@ -131,3 +131,60 @@ def normalized_grid(grid_z_coordinates, grid_norm_unit):
...
+    -------
+
+    interpolated_surface: np.ndarray
+        Interpolated surface derived from set of coordinates
+        in grid of `x_range, y_range` dimensions.
+        Array of floats of shape (`n_x_bins`, `n_y_bins`)
+
+    """
+
+    interpolated_surface = np.apply_along_axis(interpolation_by_array, axis=0, arr=array_surface)
```
lilyminium, line 188: Ok, I see that you are applying `interpolation_by_array` by row, then. In that case I suggest amending the documentation of `interpolation_by_array` to specify that the input is 1-dimensional with length `n_x`, instead of `n_x, n_y`. Although I think doing 2D interpolation would be less arbitrary (choosing which axis is x and which is y is basically a random, right?)

membrane_curvature/tests/test_membrane_curvature.py
```diff
@@ -206,7 +207,38 @@ def test_get_z_surface(x_bin, y_bin, x_range, y_range, expected_surface):
...
+               [150., 150., 150.,  150.]]),
+     np.array([[150., 150, 150.,  150.],
+               [150., 140., 145.,  150.],
+               [150., 130., 140.,  150.],
+               [150., 150., 150.,  150.]])),
+    # array 3x3  with lots of nans
+    (np.array([[np.nan, np.nan, 150.],
+              [150, np.nan, 150.],
+              [np.nan, 150., np.nan]]),
+     np.full((3, 3), 150.))
```
lilyminium, line 237: Could you please add some tests with more interesting values? e.g. in the one below, the 145 number could have come from interpolating either along the x-axis or the y-axis. In order to be diagnostic of potential current and future bugs, I'd be interested in seeing one where using 1D interpolation gives different results if it's run across the x-axis, vs. the y-axis.

```python
(np.array([[150., 150,  150.,  150.],
               [150., np.nan, np.nan,  150.],
               [150., 130., 140.,  150.],
               [150., 150., 150.,  150.]]),
     np.array([[150., 150, 150.,  150.],
               [150., 140., 145.,  150.],
               [150., 130., 140.,  150.],
               [150., 150., 150.,  150.]])),
```

membrane_curvature/tests/test_membrane_curvature.py
```diff
@@ -563,9 +595,45 @@ def test_analysis_wrapping_coordinates(self, x_bin, y_bin, box_dim, dummy_array,
...
+                                   [0., 200., 150],  [100., 200., np.nan], [200., 200., np.nan],
+                                   [0., 300., 150.], [100., 300., 150.],   [200., 300., 150.]]),
+         np.full((1, 3, 4), 150.)),
+        # array 4x4 with all 120 and many nans
+        (400, 400, 4, 4, np.array([[0., 0., np.nan], [100., 0., 120.], [200., 0., 120.], [300., 0., np.nan],
+                                   [0., 100., 120.], [100., 100., np.nan], [200., 100., 120.], [300., 100., 120.],
+                                   [0., 200., 120], [100., 200., np.nan], [200., 200., np.nan], [300., 200., 120.],
+                                   [0., 300., np.nan], [100., 300., 120.], [200., 300., 120.], [300., 300., np.nan]]),
+         np.full((1, 4, 4), 120.))
+    ])
```
lilyminium, line 629: Here as well, tests where the expected values aren't all one number are needed to really check assumptions such as:

* is it interpolating along the axis I think it is
* is it interpolating in a linear way (vs. some random other polynomial)
* is it even interpolating, or just taking the most common number?
* are we sure it's interpolating independently for every frame, or is it interpolating across different frames?

etc.

I am, incidentally, somewhat surprised that coordinates are getting wrapped with `np.nan` values in them. @richardjgowers should this be happening...?

membrane_curvature/base.py
lilyminium: Do you need `np.nanmean`? Do you expect `np.nan` values in your interpolated surfaces? If not, using `np.mean` will reveal errors in case there ever are `np.nan` values.

lilyminium (changes requested):
The tests here are a good start! However I think you definitely need to vary your values (maybe by doing tests on your added data files) for robust checking. I am already reasonably certain that `self.results.average_interpolated_z_surface` etc. are not calculated the way you want them to, which wouldn't have been caught with current test cases (although there aren't actually tests for those yet -- they need some too!).

PR #65, opened 2021-08-10
Updated README

### LINKED ISSUE #54: Installation instructions
Could you please add installation instructions? Edit: typically installation instructions are given both in the root repo README.md, and in the documentation on ReadTheDocs. At the moment, as a new user, I'm not sure how to install and use your code. There have not been any releases yet, so people will have to build from source. I usually put my dependencies into another yaml file instead of re-using my test environment, because that has extra packages such as pytest. For examples, yours could look like:

```
name: membrane-curvature
channels:
  - conda-forge
  - defaults
dependencies:
  - python
  - pip
  - numpy

    # Pip-only installs
  - pip:
    - MDAnalysis==2.0.0b0
```

If you save that as `environment.yml` in `devtools/conda-envs`, building from source would look like:

```
git clone https://github.com/MDAnalysis/membrane-curvature.git
cd membrane-curvature
conda env create -f devtools/conda-envs/environment.yaml
conda activate membrane-curvature
python setup.py install
```

Once packages are created for Pypi and conda, those instructions can be added too.

ojeda-e:
## Description
This PR helps to fix #54 and gets the README ready for the first release.

Changes in this PR:

- [x] Added sections in the README as listed in #54:
- Documentation
- Features
- Installation
- Usage
- License
(not in that order)
- [x] Added a figure in docs because it looks cooler.
Edit: ~~Changed `test_env.yaml` to `requirements.yaml` (For installation instructions)~~.
- [x] Added `requirements.yaml`, ~~changes name of env from `test` to `name: membrane-curvature`.
Then it's all more consistent.~~

Thanks.


devtools/conda-envs/environment.yaml
```diff
@@ -0,0 +1,13 @@
+name: membrane-curvature
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - python
+  - pip
+  - numpy
+  - MDAnalysisTests
```
ojeda-e, line 9: I added MDAnalysisTests because the Usage includes examples uses it, so better to have them in the environment.

lilyminium: Maybe stick it in pip with 2.0.0b0, as I think it's required to match the MDAnalysis and MDAnalysisTests versions

IAlibay: I would avoid MDAnalysisTests as a default dependency, it's 100+ MB of disk space which can be a lot.

We often use MDAnalysisTests as part of the core library usage examples, but don't make folks install it by default. What I recommend here is maybe just adding an extra few lines in the setup instructions telling folks they can install MDAnalysisTests if they want to follow the usage instructions?

devtools/conda-envs/environment.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+    # Pip-only installs
+  - pip:
+    - MDAnalysis==2.0.0b0
```
IAlibay, line 13: Would `=>` work here? This would avoid having to immediately change this once 2.0 is properly out?

README.md
```diff
@@ -6,37 +6,107 @@ Membrane Curvature
...
-The MDAkit for membrane curvature analysis is part of the [Google Summer of Code][GSoC] program 
-and it is linked to a [Code of Conduct][code_of_conduct].
+Features
+--------------
+
+With MembraneCurvature you can:
+
+- Calculate mean and Gaussian curvature from MD simulations.
+- Derive 2D curvature profiles.
+- Live a happier life.
```
IAlibay, line 22: 🤣

devtools/conda-envs/environment.yaml
ojeda-e: Thanks, that's a good point. I will leave MDAtests out for now. I added the instructions to install tests but in the Documentation section, which may appear too late in the README. I will reorganize it.

devtools/conda-envs/environment.yaml
ojeda-e: Yes, added, thanks.

README.md
```diff
@@ -6,37 +6,108 @@ Membrane Curvature
...
+
+
+Installation
+--------------
+
+To install from source:
```
IAlibay, line 28: So once you do your first release I assume the aim is that this will become available via pip / conda (pinging @lilyminium @orbeckst @fiona-naughton ) - I guess these instructions can be updated after the fact?

devtools/conda-envs/environment.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - conda-forge
+  - defaults
+dependencies:
+  - python
+  - pip
+  - numpy
+
+    # Pip-only installs
+  - pip:
+    - MDAnalysis=>2.0.0b0
```
IAlibay, line 12: Might want to check that this works, I think I had it the wrong way around in my comment yesterday i.e. it probably should be `>=` instead of `=>`.

IAlibay (approved):
lgtm - just the one `>=` vs `=>` that needs checking (approving now since I don't want to block just for that)

devtools/conda-envs/environment.yaml
ojeda-e: Thanks so much for catching this!

PR #72, opened 2021-09-03
Fixed MembraneCurvature import in API docs

### LINKED ISSUE #71: Example in API docs displays incorrect import
The [example](https://github.com/MDAnalysis/membrane-curvature/blob/4bb851efdeaaeb894b7c354fd019f5f6a97079de/membrane_curvature/base.py#L98) provided in the docstrings of `base.py` displays an incorrect import:
`from MDAnalysis.analysis import MembraneCurvature`,
that should be
`from membrane_curvature.base import MembraneCurvature`.
A link that redirects to Usage after the example provided may be useful for the user.

To fix this issue:
- [x] Change import.
- [x] Add a link to redirect to specific examples.

ojeda-e:
This PR fixes #71 

Quick changes added:

- [x] Fixed `MembraneCurvature` import.
- [x] Added links to Usage and Visualization.


ojeda-e:
Hi @orbeckst Would you please review this short PR? Thanks!

membrane_curvature/base.py
```diff
@@ -110,8 +101,15 @@ class MembraneCurvature(AnalysisBase):
...
 
     The respective 2D curvature plots can be obtained using the `matplotlib`
-    package for data visualization via `imshow`. We recommend using the
-    `gaussian` interpolation.
+    package for data visualization via `contourf` or `imshow`.
+
+    For spcecific examples visit the
+    `Usage <https://membrane-curvature.readthedocs.io/en/latest/source/pages/Usage.html>`_ page.
+    Check the `Visualization
+    <https://membrane-curvature.readthedocs.io/en/latest/source/pages/Visualization.html>`_
```
orbeckst, line 109: I generally use reST links instead of implicit links to keep it more readable
```reST
visit the usage_ page. Check the visualization_ page for ...

.. _usage: https://membrane-curvature.readthedocs.io/en/latest/source/pages/Usage.htm
.. _visualization: https://membrane-curvature.readthedocs.io/en/latest/source/pages/Visualization.html
```
However, given that this is the _same_ docs, you should be using sphinx refs so that the links always work:

```reST
visit the :ref:`usage-label` page. Check the :ref:`visualization-label` page for ...
```

where the docs in `Usage.rst` contain something like
```reST
.. _usage-label:

Usage
=========================================================

In this page, you can find examples of how to use MembraneCurvature to derive
curvature profiles in three types of systems:

:ref:`membrane-only`.
```

... but you're already using refs so you know what I am talking about.

orbeckst (changes requested):
Something got messed up and I would use `:ref:` instead of inline links.

membrane_curvature/base.py
ojeda-e: Thanks, replaced links by `:ref:` for Usage and Visualization pages.

ojeda-e:
I'll ping you again @orbeckst, just in case you have some time to review this PR. :)

membrane_curvature/base.py
```diff
@@ -110,9 +109,12 @@ class MembraneCurvature(AnalysisBase):
         gaussian_curvature = mc.results.average_gaussian_curvature
 
     The respective 2D curvature plots can be obtained using the `matplotlib`
-    package for data visualization via `imshow`. We recommend using the
-    `gaussian` interpolation.
+    package for data visualization via `contourf` or `imshow`.
```
orbeckst, line 112: I would link to the matplotlib functions with intersphinx
```suggestion
    package for data visualization via :func:`~matplotlib.pyplot.contourf` or :func:`matplotlib.pyplot.`imshow`.
```
You need to check how exactly the intersphinx links need to be structured because I am guessing here.

You also need to add the intersphinx linking to your `conf.py`.

membrane_curvature/base.py
```diff
@@ -110,9 +109,12 @@ class MembraneCurvature(AnalysisBase):
...
 
+    For specific examples visit the :ref:`usage` page.
+    Check the :ref:`visualization` page for more examples to plot
+    MembraneCurvature results using `contourf`
```
orbeckst, line 116: intersphinx link

orbeckst (approved):
Looks good.

My suggestion is to use intersphinx links to matplotlib where you can: https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html

membrane_curvature/base.py
ojeda-e: Thanks, intersphinx was already there. Added the links to `contourf`, `imshow`, and while I was there also the one for `MDAnalysis.transformations.wrap.wrap`.

Issue #74, opened 2021-09-04
Update tutorial membrane-only


ojeda-e:
Update [membrane-only tutorial](https://membrane-curvature.readthedocs.io/en/latest/source/pages/Curvature_membrane-only_systems.html) after #70 is fixed. 
Plots of mean curvature should display inverted colors.

PR #75, opened 2021-09-04
Fixed sign of mean curvature

### LINKED ISSUE #70: Sign of mean curvature
Related to #69: I think the sign of the mean curvature is reversed here compared convention.

Conventionally I think we use the upward-pointing Z-axis as the membrane normal. With that, you can use [this definition of the mean curvature](https://en.wikipedia.org/wiki/Mean_curvature#Surfaces_in_3D_space) (as you do). But I think you've added an extra negative. Should that be there?

ojeda-e:
The sign in the equation of mean curvature was wrong, indeed. Thanks, @lilyminium for catching this.
The overall term used to calculate mean curvature was quite messy, and it didn't match the one I added to the docstrings. 

I added two main changes in this PR:

- [x] Reorganized equation of mean curvature (now matches docstrings).
- [x] Updated tests.

To clarify, with the change of the sign 3 tests failed: `test_analysis_mean`, `test_mean_curvature_all`, `test_mean_curvature_small`. As you can verify, the change in the expected value of _H_ is the sign, exclusively. These three tests were updated accordingly.  
Just in case, I double-checked the term of Gaussian curvature, and it looks OK.

Since this change has an impact on the first tutorial included in docs, I opened #74 to update the tutorial once this PR is approved. 
I'm very sorry for this terrible mistake, @lilyminium @orbeckst. I hope these changes fix #70. 

Please let me know if I missing anything else here.


ojeda-e:
Hi @lilyminium, would you please take a look at the changes pushed here? :) Thanks.

lilyminium:
Sorry for the delay. Looks good to me!

lilyminium (approved):

PR #89, opened 2022-01-16
Fixed plots in documentation pages (transposed arrays).

### LINKED ISSUE #88: Examples in documentation show incorrect plots.
**Context**:
The figures included in docs use either `contourf` or `imshow`.  In both cases, the resulting surface/curvature array is plotted as `(M, N)` array-like. 

From the [plt.contourf](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.contourf.html) docs:
```
Parameters
-----------

X, Y array-like, optional: 
    The coordinates of the values in Z.

    X and Y must both be 2D with the same shape as Z (e.g. created via numpy.meshgrid), 
    or they must both be 1-D such that len(X) == N is the number of columns in Z and 
    len(Y) == M is the number of rows in Z.

    If not given, they are assumed to be integer indices, i.e. X = range(N), Y = range(M).
```
Similar story with [plt.imshow](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.imshow.html). Hence, the resulting surface/curvature array should be transposed to generate the correct plot.

To illustrate the issue, here below is an example using `plt.imshow` from the [Membrane Curvature Visualization page](https://membrane-curvature.readthedocs.io/en/latest/source/pages/Visualization.html)

**Current plot**: 😞 

drawing

**Expected plot**: ☺️ 

drawing

To fix this issue, update all the figures in the visualization and tutorial pages.

ojeda-e:
## Description
This PR fixes #88.  

## Todos
- [x] fixed plots in Visualization pages.
- [x] fixed plots in tutorial CG.
- [x] fixed plots in tutorial AA.   

## Status
- [x] Ready to go

### DIFF
--- a/docs/source/pages/Visualization.rst
+++ b/docs/source/pages/Visualization.rst
@@ -11,10 +11,20 @@ Matplotlib_. Two different approaches are suggested.
 :ref:`plt.contourf`
 
 .. warning::
-      Please note that the returned arrays of surface, mean, and Gaussian curvature 
-      should be transposed when plotting via :func:`~matplotlib.pyplot.imshow` or 
-      :func:`~matplotlib.pyplot.contourf`. This is a critical step when plotting 
-      results obtained from MembraneCurvature analyses.
+      Please note that when plotting via :func:`~matplotlib.pyplot.imshow` or
+      :func:`~matplotlib.pyplot.contourf`, the orientation of the plot in the
+      final rendering is determined by the ``origin`` argument. By default,
+      in these Matplotlib_ functions, the origin is set to the ``(left, top)``
+      corner. Additionally, for an array of shape ``(M, N)``, the first index
+      ``M`` runs along the vertical axis, and the second index ``N`` runs along
+      the horizontal axis.
+      
+      For these reasons, the returned arrays of surface, mean, and Gaussian
+      curvature should be transposed, and ``origin=lower`` should be passed to
+      :func:`~matplotlib.pyplot.imshow` or :func:`~matplotlib.pyplot.contourf`
+      to generate the correct plots. This is a critical step when visualizing
+      results obtained from MembraneCurvature analyses to avoid generating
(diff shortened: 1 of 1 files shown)


ojeda-e:
I added a warning in the `visualization.rst` page and the two tutorials. All plots included in these files were updated. NGL widgets are functional and working as expected.

To clarify why I ran this tutorial with MDAnalysis 2.1.0 and Python 3.7: I find an error when building docs with in Python 3.8+. I am working with v4.3.2 and higher. The error reads as 
```
Notebook error:
JSONDecodeError in cphtml/doctrees/nbsphinx/source/pages/Curvature_membrane_protein_nopr_all_atom.ipynb:
Expecting value: line 1 column 1 (char 0)
make: *** [html] Error 2
```
I tried to fix it but honestly, I couldn't find a way to build docs with Python 3.8+ and MDA 2.2.0+. I need to modernize some lines in Sphinx but that will be a separate issue and probably I will end up moving notebooks to pages with iPython.
 
@orbeckst would you please help me to review PR? Hopefully, I didn't miss anything. Thanks!

docs/source/pages/Visualization.rst
```diff
@@ -10,6 +10,11 @@ Matplotlib_. Two different approaches are suggested.
 
 :ref:`plt.contourf`
 
+.. warning::
```
hmacdope, line 13: I think it might be good to state the *reason* that they need to be transposed as that may make this a bit clearer?

hmacdope (changes requested):
Quick question

docs/source/pages/Visualization.rst
ojeda-e: Thanks, this is a good point. I extended the warning to add a detailed explanation of what's going on.

hmacdope (approved):
Looks great @ojeda-e. Go ahead when ready.

Issue #86, opened 2021-10-19
Warning messages significantly affect performance


ojeda-e:
Unfortunately, the more explicit warning messages introduced in  #83 significantly affect the performance of the MembraneCurvature, more particularly in membrane-protein systems, where the messages are overwhelming. 
While working on #84, I found a dramatic difference between version 0.0.1 (no explicit messages) and 0.0.2 (explicit messages). 
To note, these messages are popping up more often in membrane-protein messages, which is expected, given the fit trans+rot transformation that put lots of atoms out of boundaries.
 

Possible solutions are:
- Go back to the previous minimal messages in surface.py.
- Set a limit on the number of warning messages that can be triggered.
- Deal with the problem of grid size / wrap coordinates (probably related to #47 )

ojeda-e:
Closing this issue as the solution was implemented in #95 

For the record, this solution was tested in a system with 2500 frames and ~18900 atoms. 
Performance with explicit warning messages: 57.38 seconds 
Performance with `WarnOnce`: 5.24 seconds.

PR #83, opened 2021-10-01
Added logger to surface warning messages.

ojeda-e:
Fixes #79 

Changes include:
- [x] Added logger to warning messages in `surface.py` when atoms fall outside boundaries.
- [x] Updated message to provide more informations.
- [x] Added tests to raise warning messages using negative coordinates.
- [x] Added tests to raise warning messages using (too large) positive coordinates.

Thanks @orbeckst for pointing this out in previous reviews. Tagging also @lilyminium in case she's available to check this PR.

To start, will assign @hmacdope as a reviewer.

Thanks!

PR #95, opened 2022-10-13
Trigger explicit warning when out of boundaries only once 

ojeda-e:
## Description
As described in #86, the explicit warning messages introduced in #83 significantly affect performance. In this PR I introduced the `WarnOnce` class in `surface.py`, which provides an explicit message only once and after that triggers a generic message. It saves time when computing in membrane-protein systems, in particular. Tests were updated with regex that match both patterns.

## Status
- [x] Ready to go

### DIFF
--- a/docs/api/surface.rst
+++ b/docs/api/surface.rst
@@ -1,3 +1,4 @@
 
 .. automodule:: membrane_curvature.surface
     :members:
+    :exclude-members: WarnOnce
--- a/membrane_curvature/surface.py
+++ b/membrane_curvature/surface.py
@@ -24,6 +24,26 @@ MDAnalysis.start_logging()
 logger = logging.getLogger("MDAnalysis.MDAKit.membrane_curvature")
 
 
+class WarnOnce:
+    """
+    Class to warn atoms out of grid boundaries only once with full message.
+    After the first ocurrance, message will be generic.
+    """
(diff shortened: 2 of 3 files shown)


ojeda-e:
@hmacdope maybe you can help me with a review here, please?

hmacdope (approved):
Looks good to me @ojeda-e! Just out of curiosity what was the performance difference (roughly)?

hmacdope (approved):

hmacdope:
@ojeda-e I don't have write access so you can just approve and merge yourself.

ojeda-e:
Thanks, @hmacdope! 

> Looks good to me @ojeda-e! Just out of curiosity what was the performance difference (roughly)?

I had a membrane-protein system with a 2500-frame trajectory and ~18900 atoms. Having explicit warning messages takes 57.38 seconds to iterate over the full number of frames, vs. 5.24 with the solution here implemented in WarnOnce. I understand it's just a particular case, but still, it is a significant difference. 
This solution saves quite a lot of time with the logger/print operation, which was too expensive with the explicit warning.

> @ojeda-e I don't have write access so you can just approve and merge yourself.

Thanks for letting me know, I just added @MDAnalysis/gsoc-mentors as admin in the repo, it gave you write access right away.

PR #103, opened 2022-10-19
Add py311 to CI and towards NEP29

### LINKED ISSUE #101: Add 3.11 to CI
While waiting for the python 3.11 release, update CI.
Reference:  
 > > Just FYI, if you're about to do a release - 3.11 is just around the corner, I'd advise waiting for it.
> 
> Thanks for pointing this out @IAlibay 
> I will wait for the release and meanwhile I can create an issue to ` '3.11.0-alpha` to the CI, if that''s ok.

3.11.0-rc.2 is probably what I'd aim for here. You'll need to use a pure python action rather than mamba (see mda_xdrlib or gh-ci-cron in core).

_Originally posted by @IAlibay in https://github.com/MDAnalysis/membrane-curvature/issues/100#issuecomment-1284417305_

ojeda-e:
## Description
Fixes #101 

## Changes
- [x] Bumped actions version in CI.
- [x] Bumped python version in `minimum_requirements` CI.
- [x] Bumped numpy and MDA versions in `setup.py`.
- [x] Bumped `python_requires` to 3.8 in `setup.py`.
- [x] Added `upstream_develop` job in CI for Python 3.10 and 3.11.    

## Status
- [x] Ready to go

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -95,6 +95,10 @@ jobs:
         pip install git+https://github.com/MDAnalysis/mdanalysis.git@develop#subdirectory=package
         python -m pip install .
 
+    - name: install pytest
+      run: |
+        pip install pytest
+
     - name: run tests
       run: |
-        pytest -v --cov=membrane_curvature --cov-report=xml --color=yes membrane_curvature/tests/
+        pytest --disable-pytest-warnings membrane_curvature/tests/
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -33,9 +33,6 @@ jobs:
             os: ubuntu-latest
             python-version: 3.7
             env_file: min
-          # - os: ubuntu-latest
-          #   python-version: '3.11.0-rc.1'
-
 
     steps:
     - uses: actions/checkout@v3
@@ -80,3 +77,23 @@ jobs:
         name: codecov-${{ matrix.os }}-py${{ matrix.python-version }}
         fail_ci_if_error: True
 
+  latest_python:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        latest-python : ["3.11.0-rc.2"]
+    
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.latest-python }}
+
+    - name:  pip install membrane-curvature
+      run: |
+        python -m pip install . 
+
+    - name: run tests
+      run: |
+        pytest -v --cov=membrane_curvature --cov-report=xml --color=yes membrane_curvature/tests/
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -26,12 +26,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [macOS-latest, ubuntu-latest, windows-latest]
-        python-version: [3.7, 3.8, 3.9, "3.10"]
+        python-version: [3.8, 3.9, "3.10"]
         env_file: [env, ]
         include:
           - name: minimum_requirements
             os: ubuntu-latest
-            python-version: 3.7
+            python-version: 3.8
             env_file: min
 
     steps:
@@ -77,12 +77,12 @@ jobs:
         name: codecov-${{ matrix.os }}-py${{ matrix.python-version }}
         fail_ci_if_error: True
 
-  latest_python:
+  upstream_develop:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        latest-python : ["3.11.0-rc.2"]
+        latest-python : ["3.10", "3.11"]
     
     steps:
     - uses: actions/checkout@v3
@@ -97,7 +97,7 @@ jobs:
 
     - name: install pytest
       run: |
-        pip install pytest
+        python -m pip install pytest
 
     - name: run tests
       run: |
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -92,6 +92,7 @@ jobs:
 
     - name:  pip install membrane-curvature
       run: |
+        pip install git+https://github.com/MDAnalysis/mdanalysis.git@develop
         python -m pip install . 
 
     - name: run tests
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -31,11 +31,14 @@ jobs:
         include:
           - name: minimum_requirements
             os: ubuntu-latest
-            python-version: 3.6
+            python-version: 3.7
             env_file: min
+          # - os: ubuntu-20.04
+          #   python-version: '3.11.0-rc.1'
+
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - name: Additional info about the build
       shell: bash
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -33,8 +33,8 @@ jobs:
             os: ubuntu-latest
             python-version: 3.7
             env_file: min
-          # - os: ubuntu-20.04
-          #   python-version: '3.11.0-rc.1'
+          - os: ubuntu-latest
+            python-version: '3.11.0-rc.1'
 
 
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/CI.yaml
```diff
@@ -70,10 +70,30 @@ jobs:
...
+    
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.latest-python }}
+
+    - name:  pip install membrane-curvature
+      run: |
+        python -m pip install .
```
IAlibay, line 95: I always forget MDA is an upstream dependency. You're getting the failure because our pyproject.toml for 2.3.0 doesn't allow for a numpy variant that isn't in the 3.7-3.10 range. This does mean you probably won't be able to fully support py3.11 until MDA 2.4.0 gets released in a couple of month, but it would be a good idea for this job runner to essentially be the "cutting edge" job. i.e. installing the latest Python + the develop version of MDAnalysis. This would give you more certainty in the medium term longevity of the upcoming release (i.e. up until release the code worked with mda 2.4.0-dev0 and 3.11, so it would probably be fine when both get fully released).

To do that I would probably do:

```suggestion
        pip install git+https://github.com/MDAnalysis/mdanalysis.git@develop

        python -m pip install . 
```

ojeda-e: Thanks, I was trying to find a version of MDA with Python 3.11 to install here. Just for future reference, where can I find the list of mda `X.Y.Z-dev` releases?

ojeda-e:
@IAlibay can I have a review here, please?

.github/workflows/CI.yaml
tylerjereddy: For PyPI maybe you mean: https://pypi.org/project/MDAnalysis/#files

I'm not up to date on if we do nightly/weekly wheel builds as well.

IAlibay: We don't unfortunately (yet? I'm still trying to think of ways not to annoy pypi / conda-forge with 40+ MB of tests every night/week).

.github/workflows/CI.yaml
```diff
@@ -70,10 +70,35 @@ jobs:
...
+    
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.latest-python }}
+
+    - name:  pip install membrane-curvature
+      run: |
+        pip install git+https://github.com/MDAnalysis/mdanalysis.git@develop#subdirectory=package
```
tylerjereddy, line 95: The CI job seems to be doing the right thing at least, and a bleeding edge job like this probably makes sense anyway.

I think you should also be able to override any restrictions in `pyproject.toml` using `pip install --no-build-isolation ..` and providing the deps manually. If you wanted to use the stable source release of MDAnalysis that might be an option, but what you have here now probably makes more sense.

.github/workflows/CI.yaml
tylerjereddy: For ~weekly, you can do something like NumPy/SciPy and use anaconda.org to host a pypi mirror like: https://anaconda.org/scipy-wheels-nightly/scipy

I think we just purge out the stale files once in a while, but apart from that you can use "normal" `pip` commands directed at the pypi mirror w/ `pip install -i <address of mirror>`. I know projects do that for cross-testing for breaks in dev.

It takes time to setup of course, but there's probably some stuff that can be copy-pasted from upstream for the uploads, apart from making our own MDA upload tokens for the mirror and storing them as secrets in the CI service or whatever.

.github/workflows/CI.yaml
```diff
@@ -70,10 +70,35 @@ jobs:
...
     - name: CodeCov
-      uses: codecov/codecov-action@v1
+      uses: codecov/codecov-action@v2
       with:
         file: ./coverage.xml
         flags: unittests
         name: codecov-${{ matrix.os }}-py${{ matrix.python-version }}
         fail_ci_if_error: True
 
+  latest_python:
```
IAlibay, line 80: You're really doing a "latest everything" here (since it's also the only place you're checking against MDA develop [as far as I'm aware?]). So a name like "upstream_develop" or something might be more appropriate?

.github/workflows/CI.yaml
```diff
@@ -70,10 +70,35 @@ jobs:
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        latest-python : ["3.11.0-rc.2"]
```
IAlibay, line 85: If you're only intending to use one version using a matrix might not be worth it here.
In any case py3.11 is now officially released.
```suggestion
        latest-python : ["3.11"]
```

setup.py
```diff
@@ -47,8 +47,8 @@
     # Additional entries you may want simply uncomment the lines you want and fill in the data
     # url='http://www.my_package.com',  # Website
     install_requires=[
-        'numpy>=1.16.0',
-        'mdanalysis>=2.0.0b0',
+        'numpy>=1.20.0',
+        'mdanalysis>=2.3.0'
```
IAlibay, line 51: Any reason for being so up to date here? i.e. is there actually an API change that prevents users from using >2.0?

setup.py
```diff
@@ -59,7 +59,7 @@
     #            'Mac OS-X',
     #            'Unix',
     #            'Windows'],            # Valid platforms your code works on, adjust to your flavor
-    python_requires=">=3.6",          # Python version restrictions
+    python_requires=">=3.7",          # Python version restrictions
```
IAlibay, line 62: I'd suggest bumping this up to 3.8 (i.e. NEP29).

.github/workflows/CI.yaml
```diff
@@ -70,10 +70,35 @@ jobs:
...
+        python-version: ${{ matrix.latest-python }}
+
+    - name:  pip install membrane-curvature
+      run: |
+        pip install git+https://github.com/MDAnalysis/mdanalysis.git@develop#subdirectory=package
+        python -m pip install .
+
+    - name: install pytest
+      run: |
+        pip install pytest
```
IAlibay, line 100:
```suggestion
        python -m pip install pytest
```

IAlibay (changes requested):
Also I'd add 3.11 to the classifiers list

.github/workflows/CI.yaml
ojeda-e: Thanks for catching this.

.github/workflows/CI.yaml
ojeda-e: Yes, that's a good point. Updated.

setup.py
ojeda-e: Because of NEP29, MC runs with Python 3.8+ and numpy>1.20. Shouldn't I set the version of MDA that matches numpy>1.20 and min Python 3.8? Otherwise, how should I choose it? If I choose MDA==2.1.0, for example, wouldn't it defeat the point of NEP29?

IAlibay: I'm not sure I understand, this would be `>=<version number>`. Since 2.0.0+ supports py3.8 and numpy>=1.20, wouldn't this be fine?

ojeda-e: Thanks for clarifying. I thought I should get the version that supports _min_ numpy>=1.20 and was following the release notes for MDA 2.3.0 that say "the _minimum_ NumPy version to 1.20.0.".  Set to MDA>=2.0.0 .

ojeda-e:
Thanks for your help and patience here @IAlibay. It is ready for review.

.github/workflows/CI.yaml
```diff
@@ -70,10 +70,35 @@ jobs:
...
         flags: unittests
         name: codecov-${{ matrix.os }}-py${{ matrix.python-version }}
         fail_ci_if_error: True
 
+  upstream_develop:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        latest-python : ["3.10", "3.11"]
```
IAlibay, line 85: The word "latest" seems a bit odd here, but it's not exactly a big deal.

IAlibay (approved):
lgtm

The build times seem a bit long (~ 3-5 mins vs 1s of tests), it might be worth playing around with the `mamba` options for `conda-incubator/setup-miniconda` to see if that can be improved.

IAlibay:
> it might be worth playing around with the mamba options for conda-incubator/setup-miniconda to see if that can be improved.

Possibly a separate issue, but this might actually be one of those few cases where just using `pip` to install everything could be faster? (i.e. the benefits of conda might not be that worth it)

PR #112, opened 2023-10-27
FIx CI

IAlibay:
## Description
* Fixes codecov failing cron CI because of being called too many times
* Adds 3.11 CI
* Adds 3.12 to upstream CI
* Drop Python 3.8
* Add all necessary bits to support Python 3.12 ahead of next MDA release.
* Fix RTD builds
* Switch CI conda install to micromamba

## Status
- [x] Ready to go


IAlibay:
🤦🏽‍♂️ I accidentally pushed the update to versioneer to `main` rather than this branch.

It's effectively just the outcome of calling `versioneer install` with 0.29, and would have been included in this PR instead.

I hope this is reasonable? Otherwise I'll try to see if I can remove that commit.

IAlibay:
I've run out of velocity in trying to fix this for now - there's something breaking both rtd and py3.9, not sure why right now.

IAlibay:
Ok this shoudl be good to go.

.github/workflows/CI.yaml
```diff
@@ -44,22 +44,21 @@ jobs:
...
+          python=${{ matrix.python-version }}
+
     - name: Install package
       # conda setup requires this special shell
       shell: bash -l {0}
       run: |
         python -m pip install . --no-deps
-        conda list
+        pip list
+        micromamba list
```
RMeli, line 61:
```suggestion
```
Is this necessary? As far as I can tell, `micromamba list` is run for the `setup micromamba` step already.

IAlibay: I'm keeping the spirit of what the original CI file did (there was a conda list call here).

Generally having a step post install that tells you everything in the environment is a good idea though - it tells you if anything has changed through pip override

RMeli: I see, good point!

RMeli (approved):

IAlibay:
Thanks for the review @RMeli - looks like we aren't set as maintainers for this repo though :( 

Indeed who is meant to maintain this repo is unclear to me - this is a discussion we will need to have with @ojeda-e.

RMeli:
Always sad to see a grey check. =P Indeed, that is a discussion to be had with @ojeda-e and the GSoC mentors.

IAlibay:
There's a deeper conversation we need to have about what we (coredevs) actually maintain in the org.

IAlibay:
We will merge this without defined maintainer review to keep CI green. Thanks @RMeli for reviewing.

PR #116, opened 2023-11-20
Compute gradients with respect to grid spacing

### LINKED ISSUE #115: Gradient calculation should respect arbitrary grid spacing
The evaluations of derivatives of height field on the grid currently assume unit grid spacing without enforcement. The computed curvatures will be incorrect if the actual grid spacing isn't  one.

ctlee:
## Description
The calculation of the gradients of the height field should respect the grid spacing. This PR adds support for variable grid spacing. It also adds some tests against parametric surfaces.

fixes #115 

## Todos
Notable points that this PR has either accomplished or will accomplish.
  - [x] Make `np.gradient` respect grid spacing
  - [x] Add test to compare computed mean and gaussian curvature  against analytical formula for hemisphere
  - [x] Add test to compare computed mean and gaussian curvature  against analytical formula for monkey saddle
  - [x] Fix unit tests to check against curvature calculations which respect grid spacing. 

## Questions
N/A

## Status
- [x] Ready to go


IAlibay:
@ctlee Apologies for the very late response here - unfortunately membrane curvature doesn't have an active maintainer currently so there haven't been anyone to shephred pull requests.

We will try to get someone to review your code as soon as possible. Pinging @MDAnalysis/coredevs here in case someone might want to take on an initial review to help this process along.

membrane_curvature/tests/test_membrane_curvature.py
```diff
@@ -121,14 +121,94 @@
...
     sel = u.select_atoms('name PO4')
     return sel
 
 
+def parametric_hemisphere(
+    radius: float, nx: int, ny: int, scale: float = 5
+) -> tuple[np.ndarray, float, float]:
+    """Generate the height field and analytical mean/gaussian curvature of a hemisphere given a sampling range. The range of the patch is defined by the hemisphere radius and a scale.
+
+    Args:
```
hmacdope, line 136: Can you use numpy syntax here :) sorry to be a pain 
eg. 
```
Parameters
----------------
...
Returns 
----------

File: membrane_curvature/tests/test_membrane_curvature.py [hunk 2]
```diff
@@ -121,14 +121,94 @@
...
+
+def test_parametric_saddle_curvature():
+    for nx in [7, 23, 47, 67]:
+        for ny in [7, 23, 47, 67]:
+            height_field, K, H, dx, dy = parametric_saddle(
+                length=10, nx=nx, ny=ny, scale=800
+            )
+            mean = mean_curvature(height_field, dx, dy)
+            gauss = gaussian_curvature(height_field, dx, dy)
+
```
hmacdope, line 206: Can also check that has 0 gaussian curvature at origin?

membrane_curvature/tests/test_membrane_curvature.py
```diff
@@ -401,23 +481,23 @@ def test_analysis_get_z_surface_wrap_xy(self, universe_dummy_wrap_xy, dummy_surf
     # test mean curvature
     def test_analysis_mean_wrap(self, curvature_unwrapped_universe, dummy_surface):
         avg_mean = curvature_unwrapped_universe.results.average_mean
-        expected_mean = mean_curvature(dummy_surface)
+        expected_mean = mean_curvature(dummy_surface, curvature_unwrapped_universe.dy, curvature_unwrapped_universe.dx)
```
hmacdope, line 484: Parametrize over these arguments, to check that the original implementation still functions as intended. 

```
@pytest.mark.parametrize("varargs", [, [], [dy, dx]])
``
Or if the parametrization is too complicated split into two tests.

File: membrane_curvature/tests/test_membrane_curvature.py [hunk 4]
```diff
@@ -401,23 +481,23 @@ def test_analysis_get_z_surface_wrap_xy(self, universe_dummy_wrap_xy, dummy_surf
...
         assert_almost_equal(avg_mean, expected_mean)
 
     def test_analysis_mean_wrap_xy(self, curvature_unwrapped_universe, dummy_surface):
         avg_mean = curvature_unwrapped_universe.results.average_mean
-        expected_mean = mean_curvature(dummy_surface)
+        expected_mean = mean_curvature(dummy_surface, curvature_unwrapped_universe.dy, curvature_unwrapped_universe.dx)
```
hmacdope, line 489: Same here.

membrane_curvature/tests/test_membrane_curvature.py
```diff
@@ -401,23 +481,23 @@ def test_analysis_get_z_surface_wrap_xy(self, universe_dummy_wrap_xy, dummy_surf
...
         assert_almost_equal(avg_mean, expected_mean)
 
     # test gaussian
     def test_analysis_gaussian_wrap(self, curvature_unwrapped_universe, dummy_surface):
         avg_gaussian = curvature_unwrapped_universe.results.average_gaussian
-        expected_gaussian = gaussian_curvature(dummy_surface)
+        expected_gaussian = gaussian_curvature(dummy_surface, curvature_unwrapped_universe.dy, curvature_unwrapped_universe.dx)
```
hmacdope, line 495: Same here.

hmacdope (changes requested):
Initial review, :)

You may need to investigate some changes to docs, but otherwise great start,

IAlibay:
Updating against `main` _should_ help get CI working properly here (since we've now merged #112 )

membrane_curvature/tests/test_membrane_curvature.py
ctlee: The 0 gaussian curvature at origin condition is implicitly checked by conditions where nx and ny are odd. If the discretization doesn't include (0, 0) then we'd have to interpolate or skip the test. I think comparison against the analytical formula at corresponding grid points should be robust and complete enough.

membrane_curvature/tests/test_membrane_curvature.py
ctlee: Not sure what to do here. The original implementation is fundamentally flawed. 

The test fixture `universe_dummy_wrap` has a specific grid spacing and system dimension. Taking derivatives with respect to the default `dx`, `dy` value (=1) is like saying:

$f'(x) = \frac{f(x+h) -f(x)}{1}$.

When the derivatives must be taken with respect to the spacing, $h$:

$f'(x) = \frac{f(x+h) -f(x)}{h}$.

To have `dx`, `dy` as parameters, the fixture would have to be changed to produce meshes with varied discretization.

membrane_curvature/tests/test_membrane_curvature.py
ctlee: Same as before

membrane_curvature/tests/test_membrane_curvature.py
ctlee: Same as before

ctlee (commented):
Thanks @hmacdope  for the initial review and @IAlibay for moving this along and fixing CI. I've added a commit to change docs to numpy style. 

Please advise on what to do wrt the other requested changes.

hmacdope:
Sounds reasonble @ctlee, I am happy to give my approval on a purely code basis, but from a scientific basis I will have to take your assertions about the correctness of the new implementation relative to the old one on trust. I will tag @ojeda-e as the principal developer as she may have a view here. I will leave my view blocking for now pending waiting for feedback. :)

ctlee:
Following up on this as new versions appear to be released! Folks will get incorrect curvature values if their grid spacing isn't 1. Can verify this issue in the current implementation by comparing with the parametric surface tests added in this PR. 

An alternative approach is to add a check that the dx, dy is 1 and to throw an error otherwise. @ojeda-e

membrane_curvature/tests/test_membrane_curvature.py
```diff
@@ -129,6 +129,104 @@ def small_grofile():
...
+
+def test_parametric_hemisphere_curvature():
+    for radius in np.arange(5, 100, 5):
+        for nx in [7, 23, 47, 67]:
+            for ny in [7, 23, 47, 67]:
+                # Ensure that patch_scale is sufficiently large such that points are closely spaced
+                height_field, H, K, dx, dy = parametric_hemisphere(
+                    radius=radius, nx=nx, ny=ny, scale=200
+                )
+                mean = mean_curvature(height_field, dx, dy)
```
ojeda-e, line 172: Here `dx`  and `dy` are provided in a different order than what is defined in   [membrane_curvature/base.py](https://github.com/MDAnalysis/membrane-curvature/pull/116/files#diff-f8ce9f8ff47f05c29737ef8761516649aae370179a681477e6c37421e569a27b) . I'm failing to see why the order is different. Could you please clarify or document why is this the case?

membrane_curvature/curvature.py
```diff
@@ -46,7 +46,7 @@
 import numpy as np
 
 
-def gaussian_curvature(Z):
+def gaussian_curvature(Z, *varargs):
```
ojeda-e, line 49: I am wondering why this is `*varargs` and not `dy` `dx`. 
Since there are no tests for `*varargs`, if we keep it maybe we should add some non-uniform tests. Otherwise, maybe it would be better to have `gaussian_curvature(Z, dx=1, dy=1)`?

ojeda-e (commented):
Hi @ctlee 
Thanks for working on this changes! I left two question below to clarify my understanding. 
Would it be possible to offer an example where`*varargs` is  illustrated? It would be a nice addition to the docs.

membrane_curvature/tests/test_membrane_curvature.py
ctlee: I got confused by the ordering of the output of np.gradient calls in curvature.py. 

```python
dy, dx = np.gradient(Z)
```

suggesting that the y-axis is the first axis etc... The first axis is I think the x-axis. It's a bit weird but it seems due to row-major ordering numpy arrays are contiguous in y but not x. I've gone ahead and swapped the order of outputs of np.gradient in a new commit.

membrane_curvature/curvature.py
ctlee: I followed the prototype from [numpy.gradient](https://numpy.org/doc/stable/reference/generated/numpy.gradient.html). I think having keyword args dx, and dy are a good idea. I am however opposed to setting a default dx, dy value to be 1. I don't think that's a sensible default, most datasets won't be sampled at 1 Angstrom spacing, and folks can make mistakes by accident. WDYT?

Non-uniform spacing tests are also a good idea. I don't have time now to add that to this PR now but can commit to it in a week or two. I do know that some folks are using this library and I think it would be good to merge this PR sooner rather than later. I hope that the test won't delay that.

hmacdope:
@ojeda-e are we happy with this PR as is? I would be happy to go ahead, but thought would check with you?

ojeda-e (approved):
Hi @ctlee @hmacdope,
thanks for following up and pinging me here. Yes, I think the changes are fine. I am approving while codecov is fixed (or removed) in the CI.

hmacdope (approved):

PR #132, opened 2024-07-18
update conda env files

orbeckst:
## Description
The conda env files that are being used for CI appeared outdated. In this PR I am trying out if they can be simplified.

## Todos
Notable points that this PR has either accomplished or will accomplish.
- [x] only use conda-forge channel (removed defaults)
- [x] install MDAnalysis via conda instead of pip

## Questions
- [ ] Was there a good reason why MDAnalysis was pip installed??

## Status
- [x] Ready to go

### DIFF
--- a/devtools/conda-envs/test_min.yaml
+++ b/devtools/conda-envs/test_min.yaml
@@ -1,7 +1,6 @@
 name: test
 channels:
   - conda-forge
-  - defaults
 dependencies:
   - python
   - pip
@@ -11,7 +10,4 @@ dependencies:
   - pytest
   - pytest-cov
   - codecov
-
-    # Pip-only installs
-  - pip:
-    - MDAnalysis==2.0.0
+  - MDAnalysis==2.0.0
--- a/devtools/conda-envs/environment.yaml
+++ b/devtools/conda-envs/environment.yaml
@@ -1,12 +1,8 @@
 name: membrane-curvature
 channels:
   - conda-forge
-  - defaults
 dependencies:
   - python
   - pip
   - numpy
-
-    # Pip-only installs
-  - pip:
-    - MDAnalysis>=2.0.0b0
+  - MDAnalysis>=2.0.0
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -1,7 +1,6 @@
 name: test
 channels:
   - conda-forge
-  - defaults
 dependencies:
   - python
   - pip
@@ -11,7 +10,4 @@ dependencies:
   - pytest
   - pytest-cov
   - codecov
-
-    # Pip-only installs
-  - pip:
-    - MDAnalysis>=2.0.0
+  - MDAnalysis>=2.0.0


orbeckst:
Why the f*** is the Ubuntu 3.10 job https://github.com/MDAnalysis/membrane-curvature/actions/runs/9986845992/job/27600076214?pr=132 installing MDA 2.1.0
```
+ mdanalysis                            2.1.0  py310hd8f1fbe_1          conda-forge       2MB
```
even though the spec is "MDAnalysis[>=2.0.0]": shouldn't it just go and find 2.7.0?

It then dies with
```
MDAnalysis/lib/_cutil.pyx:1: in init MDAnalysis.lib._cutil
    ???
E   ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```
which is not surprising, given that we also installed numpy 2.0.0
```
    + numpy                                 2.0.0  py310h515e003_0          conda-forge       8MB
```
and apparently our MDA 2.1 cf package did not restrict the numpy versions on the upper end.

IAlibay:
Makes sense, the resolution order will be to pick up the highest version of numpy and the lowest version of MDAnalysis.

IAlibay:
I pushed some changes, i.e. removed `numpy` as an explicit dependency. NumPy is a dependency of MDAnalysis, so there's no point in explicitly listing it. This should fix the dependency resolution issue.

This does however point to the fact that if someone wants to install MDA right now, they'll probably encounter issues if they are doing a pure install with numpy (or anything upstream that has removed its MDA pins).

devtools/conda-envs/test_min.yaml
```diff
@@ -1,17 +1,13 @@
 name: test
 channels:
   - conda-forge
-  - defaults
 dependencies:
   - python
   - pip
-  - numpy==1.20.0
+  - numpy==1.21.0
```
IAlibay, line 7: I would suggest just removing 3.9 and bumping this up - but if folks are intent on keeping 3.9, then at least increasing to 1.21.0 is a need (since osx-arm64 will require 1.21+ for py39).

The main reason we're bumping here is because newer scipy will _work_ with older numpy, but features are mostly turned off (like pkdtree).

orbeckst: I *think* 3.9 is included here because it’s the minimally supported version and that’s what this test is supposed to check.

IAlibay: Sorry what I meant here was "maybe the minimum should be Python 3.10 not 3.9"

orbeckst:
Thanks @IAlibay — now all tests pass again 🎉 

I’d leave the ‘min’ test for now unless people think it’s useless. 

We do need an approving review to merge.

IAlibay (approved):

devtools/conda-envs/test_min.yaml
orbeckst: If you want to bump the minimally supported Python version of membrane-curvature to 3.10 then we can do that  (even though its dependencies still support 3.9). If we bump then we should do a new release with advertised 3.10 - 3.12 (not sure if 3.12 is in pyproject.toml).

orbeckst: It is in there — so it would just be removing 3.9 along spec0 Blabla.

orbeckst:
I am going to merge for right now and we can change the minimum supported version in another PR.

Issue #136, opened 2026-04-18
Build: Package update published to PyPI does not reflect latest source code changes


ojeda-e:
After publishing the last version in to PyPI, the installed package does not reflect the latest source code changes in base.py:

https://github.com/MDAnalysis/membrane-curvature/blame/cd6c1a6c658df5a6192c9bea8590ba73b1181187/membrane_curvature/base.py#L138-L140

https://github.com/MDAnalysis/membrane-curvature/blame/cd6c1a6c658df5a6192c9bea8590ba73b1181187/membrane_curvature/base.py#L188-L193

The version number incremented, but the code retrieved via `pip install` appears to be the same one as in 1.0.1.

## Expected Behavior:
The installed package should include the most recent source code changes corresponding to the new version.

## Actual Behavior:
The installed package appears to contain older code, not reflecting updates.

PR #137, opened 2026-04-18
Build: update deployment workflow w/ testpypi and pypi

ojeda-e:
While trying to fix #136 I came across an issue with the deployment workflow. This PR aims to fix it.

## Description
Updates `.github/workflows/deploy.yaml` to deploy to PyPI using a step to test on TestPyPI. All other previous changes were migrated to PR #141 

## Todos
- [x] Updated checkout to `v6`, 
- [x] Replaced `MDAnalysis/pypi-deployment` with latest version of `gh-action-pypi-publish`, `v1.14.0` following suggestion from https://github.com/MDAnalysis/GridDataFormats/blob/master/.github/workflows/deploy.yaml. *
   
* Note: 
On top of `test-built-wheel`, 4 jobs were added to the deployment workflow: `publish-testpypi`, `publish-pypi`, `verify-testpypi-install`, `verify-pypi-install`.

## Status
- [x] Ready to go

Edit2: The scope of this PR is limited to fix the deployment gh workflow ~extended quite a bit after finding CI errors~.

PR #141, opened 2026-04-23
Fix CI: update min deps, drop old Python versions, extend to py314.

ojeda-e:
(cherry picked from commit b43955ee3215213d23861112441c4f38c87c70be)

## Context:
The main blocker in #136 is adding my username to TestPyPI, which is required to properly test the deployment workflow before merging. However, this is not necessary to resolve the current CI failures in [upstream-develop 3.10](https://github.com/MDAnalysis/membrane-curvature/actions/runs/24798335966/job/72573974222?pr=140#logs).

To avoid coupling unrelated concerns, I have split the work from #136 into two separate PRs:

- One focused solely on fixing the deployment workflow (#136).
- This one, focused on fixing the CI.

This separation also helps reduce pressure on resolving the deployment-related PR immediately. IMO, while the deployment can wait, fixing the CI is more time-sensitive. I chose this approach because the CI is currently failing, and merging changes alongside unrelated deployment work would make the situation unclear and I don't feel comfortable doing so.

## Description
Note that yhe changes here are cherry-picked from #136 and are limited to what is required to restore a passing CI:

- Dropping support for older Python versions and extending support up to 3.14.
- Bumping minimum dependency versions.

### DIFF
--- a/devtools/conda-envs/test_min.yaml
+++ b/devtools/conda-envs/test_min.yaml
@@ -4,8 +4,8 @@ channels:
 dependencies:
   - python
   - pip
-  - numpy==1.23.5
-  - MDAnalysis==2.4.3
+  - numpy==1.23.0
+  - MDAnalysis==2.4.0
   - sphinx
      # Testing
   - pytest
--- a/devtools/conda-envs/test_min.yaml
+++ b/devtools/conda-envs/test_min.yaml
@@ -4,8 +4,8 @@ channels:
 dependencies:
   - python
   - pip
-  - numpy==1.23.0
-  - MDAnalysis==2.4.0
+  - numpy==1.23.5
+  - MDAnalysis==2.4.3
   - sphinx
      # Testing
   - pytest
--- a/devtools/conda-envs/test_min.yaml
+++ b/devtools/conda-envs/test_min.yaml
@@ -4,8 +4,8 @@ channels:
 dependencies:
   - python
   - pip
-  - numpy==1.21.0
-  - MDAnalysis==2.0.0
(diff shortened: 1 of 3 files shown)


devtools/conda-envs/test_min.yaml
```diff
@@ -4,8 +4,8 @@ channels:
 dependencies:
   - python
   - pip
-  - numpy==1.21.0
-  - MDAnalysis==2.0.0
+  - numpy==1.23.5
+  - MDAnalysis==2.4.3
```
RMeli: Is it correct that these do not match exactly the lower bounds in the `pyproject.toml`?
```
    'numpy>=1.23.0',
    'mdanalysis>=2.4.0'
```

RMeli (approved):
Minor question, otherwise LGTM.

devtools/conda-envs/test_min.yaml
ojeda-e: Thanks for noticing this! This is incorrect, indeed. In the original PR #137, these deps are bumped in the pyproject.toml as well. I left [this](https://github.com/MDAnalysis/membrane-curvature/pull/137/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L24-L25) behind. Updating now.

ojeda-e: I left the former
```
  - numpy==1.23.5
  - MDAnalysis==2.4.3
```
I forgot that they don't match because conda can't resolve with the matching ones in the `pyproject.toml`.

PR #146, opened 2026-04-28
Feat: Implement Fourier modes as a new strategy to derive surfaces

### LINKED ISSUE #148: Feat: add method to derive surfaces with Fourier Series.
**Feature to be added in v2.0.0**

Currently there is only one strategy to derive surfaces: from atomGroup -> grid -> surface.
To fix this issue: implement a strategy to derive surfaces with Fourier.

The API should look like:

```python
    mc_fourier = MembraneCurvature(
    universe,
    select="name PO4",
    surface_method="fourier",  # surface_method="binning" would run the current approach
    fourier_n = 2, # def value, but flexible if user wants to modify it
    fourier_m = 2, 
).run()
```

This approach would also fix #45 because the modes in the Fourier Series would account for changes in box lengths on a per-frame basis.

ojeda-e:
## Description
This PR introduces a nee feature to derive surfaces in addition to the already exisitng binning strategy. Here I am adding a different approach, in which we derive surfaces using Fourier series to fit a periodic surface with least-squares. Fixes #148 

This implementation is inspired from:

- Rhouma et al., Fourier method for large scale surface modeling and registration, Computers & Graphics (2009), doi: [10.1016/j.cag.2009.03.002](https://doi.org/10.1016/j.cag.2009.03.002).
- Leicester et al., Description of molecular surface shape using Fourier descriptors, Journal of Molecular Graphics (1988), doi: [10.1016/0263-7855(88)85008-2](https://doi.org/10.1016/0263-7855(88)85008-2).
  
## Changes in thir PR
- [x] Added code for surface derivation with Fourier with non-redundant modes in `fourier_surface.py`.
- [x] Update `base.py` to include `surface_method == 'fourier'` and `surface_method='binning'`.
- [x] Renamed former `surface.py` to `surface_binning.py`.
- [x] Added separate tests for all the new Fourier functionality in `test_fourier_surface.py`. Kept coverage at 100%.
- [x] Added min docs with references (Rhouma et al., Leicester et al.), and with public and private API.

## Notes
- The tests in `test_membrane_curvature` are lengthy and I prefer to have a separate file for maintenability and readability purposes. 
- I am aware this PR doesn't include comprehensive docs. However, it is already quite large and I'd prefer to work on a separate PR to update figures and doc pages.
- Below, I'll leave some questions for reviewers.


docs/api/fourier_surface.rst
```diff
@@ -0,0 +1,5 @@
+
+.. automodule:: membrane_curvature.fourier_surface
+   :members:
+   :private-members:
```
ojeda-e, line 4: I have a lot of private functions in this file. I am not sure if for MDAkits there is a recommendation on adding/hiding public functions. 
My opinion is that the API although private helps to illustrate the logic to derive the surface and it should be available.

membrane_curvature/tests/test_fourier_surface.py
```diff
@@ -0,0 +1,471 @@
...
+    X_rel: np.ndarray
+    Y_rel: np.ndarray
+    A00: float
+    A: float
+    coeffs: dict
+    Lx: float
+    Ly: float
+
+
+# This fixture  is a dummy universe to check Fourier surface for
```
ojeda-e: I am using a fixture with this toy grid of 2x2 that makes the logic more redable, at least from the test perspective. It also makes writing the unit tests easier.

membrane_curvature/tests/test_fourier_surface.py
```diff
@@ -0,0 +1,471 @@
...
+    _eval_fourier_surface,
+    _fourier_fit_from_atoms,
+    n_fourier_parameters,
+    fourier_mode_list,
+    fourier_height_from_atoms,
+    fourier_height_derivatives_from_atoms,
+)
+
+
+@dataclass(frozen=True)
```
ojeda-e, line 26: I am biased because I do like data classes, so added one here.  If there is any argument or rationale as to why adding this is a bad idea, I'm open to discuss and hear PoVs.

membrane_curvature/tests/test_fourier_surface.py
```diff
@@ -0,0 +1,471 @@
...
+        A00=9.0,
+        A=0.5,
+        coeffs={(1, 0): (0.5, 0.0)},
+        Lx=2.0,
+        Ly=2.0,
+    )
+
+def test_calculate_a00_2x2_grid(dummy_fourier_universe, dummy_fourier_fit):
+    """Test A_00 matches np.mean of z coords"""
+    expected_A00 = 9 # equivalent to np.mean(positions[:, 2])
```
ojeda-e: This is an np.mean. With the 2x2 grid I did it on paper first and I know it's 9. Not sure if hardcoded values in this case would make sense. Left the equivalent of numpy as acomment on the right.

membrane_curvature/fourier_surface.py
```diff
@@ -0,0 +1,685 @@
...
+        Intermediate :math:`\partial^2 h/\partial \phi^2` used for
+        ``fxx``, ``fyy``, ``fxy`` accumulation.
+    """
+    h = A * cos_phase + B * sin_phase
+    h_phi = -A * sin_phase + B * cos_phase
+    h_phiphi = -A * cos_phase - B * sin_phase
+    return h, h_phi, h_phiphi
+
+
+def _eval_fourier_surface(
```
ojeda-e, line 512: The point of `_eval_fourier_surface` is to return both surface and derivatives. Separating them doesn't make sense in this context -in contrast to the binning strategy- because the derivatives are a property of the Fourier representation, and they're calculated with the `coeffs`, `A_00`, etc. However, I though of a case in which the user could be interested in the surface only. In this case the derivatives are not needed, so I though having a bool option for derivatives could offer more flexibility to the user.

membrane_curvature/tests/test_fourier_surface.py
```diff
@@ -0,0 +1,471 @@
...
+        A00=9.0,
+        coeffs={},
+        derivatives=True,
+    )
+    assert_almost_equal(Z, np.full_like(X_rel, 9.0))
+    assert_almost_equal(fx, np.zeros_like(X_rel))
+    assert_almost_equal(fy, np.zeros_like(X_rel))
+    assert_almost_equal(fxx, np.zeros_like(X_rel))
+    assert_almost_equal(fyy, np.zeros_like(X_rel))
+    assert_almost_equal(fxy, np.zeros_like(X_rel))
```
ojeda-e, line 312: I honestly have no idea if this could be written better or in a more pythonic way. [Here](https://github.com/MDAnalysis/membrane-curvature/pull/146/changes#r3156492473) I explain why the function should return surface and derivatives. 
If there is a better way to write these assertions ina less redundant way, please let me know.

membrane_curvature/base.py
```diff
@@ -138,6 +168,27 @@ def __init__(self, universe, select='all',
...
+            n_param = n_fourier_parameters(self.fourier_m, self.fourier_n)
+            n_atom = len(self.ag)
+            if n_atom < n_param:
+                max_square_mode = max(int((np.sqrt(n_atom) - 1) // 2), 0)
+                raise ValueError(
+                    f"surface_method='fourier' needs at least {n_param} atoms in the selection "
+                    f"(fourier_m={self.fourier_m}, fourier_n={self.fourier_n}), got {n_atom}. "
+                    f"Suggested max modes for {n_atom} atoms is "
+                    f"fourier_m = fourier_n = {max_square_mode}."
+                )
```
ojeda-e, line 216: The purpose of adding this explicit error message is to account for cases where users specify an arbitrary number of modes beyond the default values (`m=n=2`). In these situations, the model may enter an overfitting regime -infite solutions-. This message indicates the min number of atoms required to safely run the specified `fourier_m` and `fourier_n` modes.

ojeda-e:
@hmacdope I tagged you as a reviewer just in case you have some spare time, pretty please 🙂.

docs/api/fourier_surface.rst
ojeda-e: @orbeckst, maybe you have suggestions or opinions here?

orbeckst: It's research software, so if in doubt I'd show more than less.

ojeda-e:
@BradyAJohnston are you available/interested in taking a look at the changes here? 👀

BradyAJohnston:
@ojeda-e should be able to have a look over the next couple of days!

ojeda-e:
After merging the changes in #151 I had to squash all commits to a single one because the rebase was painful and messy. 
@RMeli if you want to take a look at this one, your review is welcome too.

membrane_curvature/__init__.py
```diff
@@ -11,7 +11,7 @@
 #
 
 # Add imports here
-from .surface import (
+from .binning_surface import (
```
orbeckst, line 14: Renaming a module breaks the API. If you're on semantic versioning, you'll need to increase your major release number (eg 1.x -> 2.x).

The courteous thing to do would be to deprecate `surface` (and keep a stub module in place) and then remove in the next major. However, if you feel that your users will be ok with just going to the next major version then don't go through the considerable hassle of deprecations.

membrane_curvature/base.py
orbeckst: I'd add the explanation as comments.

membrane_curvature/binning_surface.py
orbeckst, line 1: see comment above: renaming a module breaks the API so requires a new major version under semantic versioning

membrane_curvature/__init__.py
orbeckst: It looks as if this is going into a 2.0.0 so that's ok then.

Perhaps add a 
```rst
.. versionchannged:: 2.0.0
   Renamed module from ``surface`` to ``binning_surface``.
```

membrane_curvature/fourier_surface.py
orbeckst: I am not a fan of changing return values based on input parameters because then I have to capture the return values as a tuple and depending on what I selected must choose what to pull out. For user-facing code a dict as a return value is a bit cleaner because the key lookup is self-documenting. 

However, if this is not a user-facing function then it barely matters.

If performance is a consideration then definitely allow yourself to not compute what is not needed.

membrane_curvature/fourier_surface.py
```diff
@@ -0,0 +1,669 @@
...
+        Mean term :math:`A_{00}`.
+    coeffs : dict[tuple[int, int], tuple[float, float]]
+        Mode amplitudes ``(A, B)`` from :func:`_fourier_fit_from_atoms` / :func:`_unpack_coefficients`.
+    derivatives : bool
+        If ``False``, return only ``Z``. If ``True``, return ``Z`` and five derivative grids
+        ``fx``, ``fy``, ``fxx``, ``fyy``, ``fxy``.
+
+    Returns
+    -------
+    ndarray or tuple of ndarray
```
orbeckst: This seems the worst alternative return to me. At least make it always a tuple so that you can always get at `Z` with `retval[0]`.

membrane_curvature/fourier_surface.py
```diff
@@ -0,0 +1,669 @@
...
+3. Build a design matrix from cosine and sine basis functions
+    - :func:`_build_fourier_matrix`
+4. Solve the least-squares problem for the Fourier coefficients
+    - :func:`_fourier_fit_from_atoms`
+    - :func:`_unpack_coefficients`
+5. Reconstruct the continuous surface on a grid:
+    - :func:`_bin_centre_mesh`
+    - :func:`_eval_fourier_surface`
+6. Evaluate partial derivatives analytically:
+    - :func:`_eval_fourier_surface` with ``derivatives=True``, or
```
orbeckst, line 33: Perhaps make all these functions public??

orbeckst (commented):
Using the Fourier surface looks like a neat idea. I don't have time to look at the science, sorry, I just left a few surface-level comments. 

You'll have to decide which functions should be public. Do you expect users to compose them and use them separately? Or are they really all internal (and you'd like the freedom to change them at any time)?

RMeli:
@ojeda-e I'll try to have a look in the coming days, but please don't wait on my review. I'm not sure if I can find enough time to review such a big PR in a reasonable timeframe.

membrane_curvature/fourier_surface.py
ojeda-e: I hesitated on this, went back and forth before submitting the PR. I ended up leaving them private.

>You'll have to decide which functions should be public. Do you expect users to compose them and use them separately? Or are they really all internal (and you'd like the freedom to change them at any time)?

The main reason I made separate functions (one per step) is to modularize the code. However, each individual function doesn't provide much value; they are just helpers to ultimately derive surfaces and calculate derivatives. As such I wouldn't expect any user to run them separately, hence why I made them private.

I think that one of advantage of leaving them private is that if I need to fix or tweak the implementation for the Fourier surface, I will be able to do it without changing the two key public entry points in `fourier_height_from_atoms` and `fourier_height_derivatives_from_atoms` which are used in `curvature.py`.

membrane_curvature/fourier_surface.py
```diff
@@ -0,0 +1,667 @@
...
+    coeffs : dict[tuple[int, int], tuple[float, float]]
+        Mapping ``(m, n) -> (A, B)`` cosine and sine coefficients per mode.
+
+    Warns
+    -----
+    UserWarning
+        If the design matrix is rank-deficient or underdetermined (same condition as
+        :func:`fourier_height_from_atoms` and :func:`fourier_height_derivatives_from_atoms`).
+    """
+    x0, y0 = x_range[0], y_range[0]
```
ojeda-e: I am not sure I should check the atom coordinates for the Fourier fit before doing least-squares, something like `if positions.ndim != 2 ` . 

Should we validate positions here? `MembraneCurvature` already rejects an empty selection, but the following functions can also be called with raw NumPy arrays:
- `fourier_height_derivatives_from_atoms`
- `fourier_height_from_atoms`
-  `_fourier_fit_from_atoms`

I'd rather avoid redundant code, but since this path isn’t always reached via MDAnalysis I wonder if it is better to add it.

membrane_curvature/fourier_surface.py
```diff
@@ -0,0 +1,667 @@
...
+    n_y_bins : int
+        Number of bins along :math:`y`.
+
+    Returns
+    -------
+    X_rel : ndarray
+        Mesh of relative :math:`x` coordinates, shape ``(n_x_bins, n_y_bins)``.
+    Y_rel : ndarray
+        Mesh of relative :math:`y` coordinates, same shape as ``X_rel``.
+    """
```
ojeda-e, line 640: I am thinking of a similar check situation here, on the lines of a fail-fast approach. If `n_x_bins` or` n_y_bins ≤ 0`, `_bin_centre_mesh` can result in empty grids or edge NumPy cases. Since the mesh is only defined for positive integer bin counts, is it better to add a check before the least-squares fit to avoid doing a fit when the grid is invalid?

Actually, this would apply to several downstream functions. And the more I think about, the more I am convinced the check should be added.

membrane_curvature/fourier_surface.py
```diff
@@ -0,0 +1,667 @@
...
+        Mean height coefficient :math:`A_{00}`.
+    coeffs : dict[tuple[int, int], tuple[float, float]]
+        Mapping ``(m, n) -> (A, B)`` cosine and sine coefficients per mode.
+
+    Warns
+    -----
+    UserWarning
+        If the design matrix is rank-deficient or underdetermined (same condition as
+        :func:`fourier_height_from_atoms` and :func:`fourier_height_derivatives_from_atoms`).
+    """
```
ojeda-e, line 400: I am not sure I should check the atom coordinates for the Fourier fit before doing least-squares, something like `if positions.ndim != 2; raise error`.

Should we validate positions here? `MembraneCurvature` already rejects an empty selection, but the following functions can also be called with raw NumPy arrays:
- `fourier_height_derivatives_from_atoms`
- `fourier_height_from_atoms`
-  `_fourier_fit_from_atoms`

I'd rather avoid redundant code, but since this path isn’t always reached via MDAnalysis I wonder if it is better to add it.

ojeda-e (commented):
@orbeckst I have two specific questions that I’d like to confirm before moving forward with the implementation (and therefore make this PR slightly longer).

membrane_curvature/fourier_surface.py
orbeckst: If you don't expect users to ever use them (or don't want to support that) then keep private.

membrane_curvature/fourier_surface.py
orbeckst: If you cannot trust the input (basically, any public function) then I'd check. On the other hand, if you know that the function is only called as a part of a specific workflow and you know that the input has already been validated then I'd include an `assert`.

orbeckst: If this is function is a good point to check then check here and let all calling code fail here (and possibly catch the exception and provide additional info).

membrane_curvature/fourier_surface.py
orbeckst: Fail-fast is good; if this is a good place for this check then I'd do it here.

orbeckst: Calling functions might still need to catch the exception to provide feedback to the user what the problem in the input parameters was (and then re-raise).

membrane_curvature/fourier_surface.py
ojeda-e: I'll keep them private and for clarity they will be included in API docs.

membrane_curvature/fourier_surface.py
ojeda-e: This is a good idea, thanks. I wasn't considering a try except but that is probably the best way to display errors and provide additional/detailed info to the user.

ojeda-e: Motivated by your comment and after pondering it for a while, I ended up implementing an approach where I have all the validators in a new file called [fourier_validators.py](https://github.com/MDAnalysis/membrane-curvature/pull/146/changes#diff-068c91eea5f00389626cc6b8903fbc6a07eb36ec98ce77392f49beb6fae676cb). Each one is single-purpose and offers a clear message to the user. Then we have reusable checks and the public entry points have the validation I was looking for before fitting.

membrane_curvature/fourier_surface.py
ojeda-e: Thanks for your input. Added with fourier validators.

PR #170, opened 2026-05-25
Accept `wrap=True` only when `surface_method='binning'`

### LINKED ISSUE #169: Enhance: wrapping should run only when `surface_method='binning'`
At the moment, the MembraneCurvature API accepts `wrap=True` when `surface_method='fourier'`. However, this is not necessary because Fourier doesn't rely on atom coordinate wrapping, instead it handles periodic xy via `np.mod`:

https://github.com/MDAnalysis/membrane-curvature/blob/6c1ed09cf1b3100f73b214e5e15aba244f2e4740/membrane_curvature/fourier_surface.py#L411-L413

Then, it's best to have wrapping as a method-aware calculation.

To fix this issue, allow `wrap` only when `surface_method='binning'`, skip it when `surface_method='fourier'`.

ojeda-e:
Fixes #169 

## Description

Now that we have another method for deriving surfaces, the original coordinate wrapping should stay for `method_surface='binning'` only. Coordinate wrapping is not required for Fourier. To avoid confusion to the users, this PR updates the logic for the coordinate wrapping, when the user passes the parameter `wrap`.

 - If `surface_method` is `'fourier'` -> 
    * it checks whether wrap is True, raises an error.
    * then forces `self.wrap = False`
- Otherwise,
    * if the user did not specify wrap (wrap is None) ->
      * use the default value True
      * otherwise -> 
         - use whatever value the user provided

- Apply wrapping=True only when `method_surface='binning'`. Set default to None (the default surface method is Fourier).
- Update docs accordingly.
- Fixed typos and clarify comments.

## PR Checklist

 - [x] Tests updated/added?


membrane_curvature/base.py
```diff
@@ -124,15 +141,39 @@ class MembraneCurvature(AnalysisBase):
     values of mean and Gaussian curvature averaged over the `n_frames` of the
     trajectory.
 
-    Raises
```
ojeda-e, line 127: The Raises section was moved before the See also section.

PR #188, opened 2026-06-01
Feat: Add brick-wall FFT filtering for binning method 

### LINKED ISSUE #187: Add low-pass filtering to binning method before finite-difference calculations
The current binning method can introduce artifacts that are amplified by finite-difference calculations, resulting in edge effects, particularly noticeable:
- near bin boundaries and at the domain edges, where finite-difference operators are most sensitive.
- in calculations of Gaussian curvature, where small binning artifacts can produce disproportionately large errors that are further amplified by  `np.gradient`.

A low-pass filtering step applied to the binned data prior to computing finite differences would help reduce numerical noise, improve derivatives, and reduce edge-related artifacts without significantly affecting the underlying curvature values.

An example of how `np.nan` can introduce artifacts through finite-difference calculations in the computation of curvature was previously discussed in #45 (See [comment](https://github.com/MDAnalysis/membrane-curvature/issues/45#issuecomment-885134394)).

To clarify, although the `'fourier'` method fixes this problem, the idea is to also fix it for `'binning'`.

ojeda-e:
Fixes #187 

**Note for potential reviewers:**
This PR may look a bit too long, but the main scientific changes are in `fft_filter.py`, with the respective updates in `base.py` to run the filter on the averaged surface, but before calculating curvature.

## Description

New feature to have an optional brick-wall FFT filter in $q$ (reciprocal space), exclusive for the binning method. This filter is applied after the average surface, and before mean and gaussian curvatures are calculated.

- Added fft_filter and updated logic in base to run filter only for binning. In base, I also rephrased some docstrings.
- Simplified logic in base, because Fourier was a bit too redundant. The logic here is clearer, IMO.
- Included tests for the fft filter, and updated tests for membrane curvature (adding `filter=None`)
- Updated doc pages: Usage and algorithm. For the latter, new figure added. 
- Changed format for sections README, fixed typos and added a quick example to run MembraneCurvature with biining+filter.

## PR Checklist

 - [x] Tests updated/added?


docs/source/pages/Usage.rst
```diff
@@ -207,10 +207,58 @@ coordinate wrapping with ``wrap=True``:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. warning::
+  FFT filtering is available only with ``surface_method='binning'``. It does **not**
+  filter each frame. 
+  
+The order of operations is:
+
+1. Per frame: bin atoms, store height field  in ``results.z_surface``.
+1. After the trajectory is processed, time-average ``z_surface`` over frames, optionally
```
tylerjereddy: Do you have two steps "1"s here?

docs/source/pages/Usage.rst
```diff
@@ -207,10 +207,58 @@ coordinate wrapping with ``wrap=True``:
...
+2.1.2.1 Binning with FFT filtering 
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. warning::
+  FFT filtering is available only with ``surface_method='binning'``. It does **not**
+  filter each frame. 
+  
+The order of operations is:
+
+1. Per frame: bin atoms, store height field  in ``results.z_surface``.
```
tylerjereddy: extra space before "in" maybe?

membrane_curvature/tests/test_fft_filtering.py
```diff
@@ -0,0 +1,272 @@
...
+    universe = mda.Universe(positions, n_atoms=9)
+    universe.dimensions = [300, 300, 300, 90.0, 90.0, 90.0]
+    return universe
+
+
+def test_resolve_fft_filter_q():
+    assert resolve_fft_filter({'q': (0.0, 0.5)}, 10.0, 10.0) == (0.0, 0.5)
+
+
+def test_resolve_fft_filter_q_list_pair():
```
tylerjereddy: minor: so similar to above you could parametrize

membrane_curvature/tests/test_fft_filtering.py
```diff
@@ -0,0 +1,272 @@
...
+            [200.0, 200.0, 150.0],
+        ]
+    )
+    universe = mda.Universe(positions, n_atoms=9)
+    universe.dimensions = [300, 300, 300, 90.0, 90.0, 90.0]
+    return universe
+
+
+def test_resolve_fft_filter_q():
+    assert resolve_fft_filter({'q': (0.0, 0.5)}, 10.0, 10.0) == (0.0, 0.5)
```
tylerjereddy: here and elsewhere--the usual caveat about exact equality checks with floats vs. `assert_allclose`

membrane_curvature/tests/test_fft_filtering.py
```diff
@@ -0,0 +1,272 @@
...
+def test_high_frequency_removed():
+    x = np.linspace(0, 2 * np.pi, 32, endpoint=False)
+    y = np.linspace(0, 2 * np.pi, 32, endpoint=False)
+    xx, yy = np.meshgrid(x, y, indexing='ij')
+    height = np.sin(3 * xx) + np.sin(12 * yy)
+    dx = dy = 2 * np.pi / 32
+
+    filtered = apply_fft_filter(height, dx, dy, (0.0, 0.15))
+    residual = height - filtered
+    assert np.std(residual) > np.std(filtered)
```
tylerjereddy: nit: both values in the comparison are a function of the output of the function under test, which is perhaps not ideal

membrane_curvature/tests/test_fft_filtering.py
```diff
@@ -0,0 +1,272 @@
...
+    ).run()
+    wrapped = universe_dummy_wrap.atoms.copy()
+    wrapped.wrap()
+    height = get_z_surface(wrapped.positions, n_x_bins=3, n_y_bins=3, x_range=x_range, y_range=y_range)
+    return mc, q_high, height
+
+
+def test_membrane_curvature_average_z_surface_is_filtered_time_mean(binning_3x3_fft_mc):
+    mc, q_high, height = binning_3x3_fft_mc
+    expected = apply_fft_filter(height, mc.dx, mc.dy, (0.0, q_high))
```
tylerjereddy: perhaps slightly suboptimal that the "actual" `mc` object from the function under test is also used to calculate expected values

membrane_curvature/tests/test_fft_filtering.py
```diff
@@ -0,0 +1,272 @@
...
+        fft_filter='auto',
+        wrap=False,
+    ).run()
+
+
+def test_membrane_curvature_average_z_surface_is_fft_of_time_mean(two_frame_binning_auto_mc):
+    mc = two_frame_binning_auto_mc
+    q_bounds = resolve_fft_filter(mc.fft_filter, mc.dx, mc.dy)
+    z_avg = np.nanmean(mc.results.z_surface, axis=0)
+    expected = apply_fft_filter(z_avg, mc.dx, mc.dy, q_bounds)
```
tylerjereddy: here again, `expected` is deriving values from `mc` "actual" output, which is perhaps not ideal from a purist standpoint; may just be a nit though

membrane_curvature/tests/test_fft_filtering.py
```diff
@@ -0,0 +1,272 @@
...
+    assert_allclose(mc.results.average_z_surface, expected, rtol=1e-10, equal_nan=True)
+
+
+def test_membrane_curvature_average_mean_from_filtered_time_average(two_frame_binning_auto_mc):
+    mc = two_frame_binning_auto_mc
+    q_bounds = resolve_fft_filter(mc.fft_filter, mc.dx, mc.dy)
+    z_avg = np.nanmean(mc.results.z_surface, axis=0)
+    z_filtered = apply_fft_filter(z_avg, mc.dx, mc.dy, q_bounds)
+    assert_allclose(mc.results.average_mean, mean_curvature(z_filtered, mc.dx, mc.dy), rtol=1e-10, equal_nan=True)
```
tylerjereddy: same here -- it may be that `mc.dx` and `mc.dy` are basically just "passed through," but still something *could* go wrong to change the expected value and mask a real problem

membrane_curvature/tests/test_fft_filtering.py
```diff
@@ -0,0 +1,272 @@
...
+    ).run()
+    assert mc.fft_filter is None
+
+
+def test_membrane_curvature_binning_default_fft_filter_auto(universe_dummy_wrap):
+    mc = MembraneCurvature(universe_dummy_wrap, n_x_bins=3, n_y_bins=3, surface_method='binning').run()
+    assert mc.fft_filter == 'auto'
+    q_nyq = nyquist_q(mc.dx, mc.dy)
+    q_bounds = resolve_fft_filter(mc.fft_filter, mc.dx, mc.dy)
+    assert q_bounds == (0.0, 0.5 * q_nyq)
```
tylerjereddy: here too I suppose

membrane_curvature/tests/test_fft_filtering.py
```diff
@@ -0,0 +1,272 @@
...
+
+
+def test_membrane_curvature_binning_fft_filter_none(universe_dummy_wrap):
+    mc = MembraneCurvature(universe_dummy_wrap, n_x_bins=3, n_y_bins=3, surface_method='binning', fft_filter=None).run()
+    assert mc.fft_filter is None
+    z_avg = np.nanmean(mc.results.z_surface, axis=0)
```
tylerjereddy: here the expected value looks like a more substantial piece of the "actual" value, which makes me a bit more nervous; are the arrays small enough to hard code?

membrane_curvature/fft_filtering.py
```diff
@@ -0,0 +1,361 @@
...
+
+import warnings
+
+import numpy as np
+
+
+def _validate_bin_widths(dx, dy):
+    """Raise if bin widths are not strictly positive."""
+    if dx <= 0.0 or dy <= 0.0:
+        raise ValueError(f'Bin widths must be positive. Got dx={dx}, dy={dy}')
```
tylerjereddy: maybe "positive nonzero" or similar?

membrane_curvature/fft_filtering.py
```diff
@@ -0,0 +1,361 @@
...
+    Warns
+    -----
+    UserWarning
+        If ``q_high`` is above what the bin grid can represent (Nyquist), or if
+        ``q_low > 0`` (the filter no longer keeps the average height).
+        Use ``q_low=0`` for a usual low-pass).
+    """
+    if not isinstance(bounds, (tuple, list)) or len(bounds) != 2:
+        raise ValueError('fft_filter["q"] must be a pair (q_low, q_high)')
+    q_low, q_high = float(bounds[0]), float(bounds[1])
```
tylerjereddy: tests still pass locally if I remove the `float` coercions -- I usually prefer to avoid type coercions unless tests prove they are needed

membrane_curvature/fft_filtering.py
```diff
@@ -0,0 +1,361 @@
...
+    if q_low == q_high:
+        raise ValueError(
+            f'fft_filter q_low and q_high must differ. Got ({q_low}, {q_high}). '
+            'An equal band removes all but modes at a single |q| (often none on the grid), '
+            'which zeros the filtered height. q_low must be <= q_high.'
+        )
+    if q_low > 0.0:
+        warnings.warn(
+            f'fft_filter q_low={q_low} removes the mean height. Use q_low=0.0 for a standard low-pass.',
+            stacklevel=3,
```
tylerjereddy, line 253: `stacklevel` can be handled automatically when your minimum supported Python is `>=3.12`; good reason to still support `3.11`?

membrane_curvature/fft_filtering.py
```diff
@@ -0,0 +1,361 @@
...
+        set to the **average height of all occupied bins** on that map (not zero, and not
+        interpolated from neighbours).
+
+        This is a simple placeholder to run the filter. It can make the surface look
+        too smooth or wavy near large gaps, especially when many bins are empty. After
+        the inverse FFT, empty bins are set back to ``NaN``. If every bin is empty, the
+        fill value is ``0.0``.
+    """
+    if not nan_mask.any():
+        return height
```
tylerjereddy, line 374: tests pass locally if I purge this check:

```diff
--- a/membrane_curvature/fft_filtering.py
+++ b/membrane_curvature/fft_filtering.py
@@ -351,8 +351,6 @@ def _height_for_fft(height, nan_mask):
         the inverse FFT, empty bins are set back to ``NaN``. If every bin is empty, the
         fill value is ``0.0``.
     """
-    if not nan_mask.any():
-        return height
```

membrane_curvature/fft_filtering.py
```diff
@@ -0,0 +1,361 @@
...
+    fill_value = np.nanmean(height)
+    if not np.isfinite(fill_value):
+        fill_value = 0.0
+    filled_height = height.copy()
```
tylerjereddy, line 378: tests passed locally if I purge the copy op:

```diff
--- a/membrane_curvature/fft_filtering.py
+++ b/membrane_curvature/fft_filtering.py
@@ -356,6 +356,6 @@ def _height_for_fft(height, nan_mask):
     fill_value = np.nanmean(height)
     if not np.isfinite(fill_value):
         fill_value = 0.0
-    filled_height = height.copy()
+    filled_height = height
```

tylerjereddy (commented):
I only had the bandwidth to give you a surface level review. In these cases what I often do is "attack" the source by randomly deleting certain sections of logic to see if the tests still pass, so you'll see some comments about that.

I'm also not a domain expert on FFTs. But anyway, maybe better than no reviewing..

Workflow nit -- I noticed that you suppressed the tests in your coverage report when reviewing this--FWIW, I typically prefer to include the tests because it helps discover test code/infrastructure that is extraneous/not running/misnamed, etc.

membrane_curvature/tests/test_fft_filtering.py
ojeda-e: good catch, yes!

docs/source/pages/Usage.rst
ojeda-e: Hm, I intended to use this option to have an ordered list without passing explicit values for the numbers (like shown here: https://www.markdownguide.org/basic-syntax/#ordered-lists) but you're right, this is not rendering as I expected. Changed to use `2.` explicitly for the second line, looks fine locally but will keep an eye on this once it renders with the pushed changes.

membrane_curvature/tests/test_fft_filtering.py
ojeda-e: Hm... yes, I think you are on point. I will update here and below.

membrane_curvature/tests/test_fft_filtering.py
ojeda-e: That's a fair concern. Yes, they are small. Updated to hardcode an np array, now the comparison is against the array.

membrane_curvature/fft_filtering.py
ojeda-e: thanks, great catch!

membrane_curvature/fft_filtering.py
ojeda-e: My only rationale is `3.11` reaches the end of life until 27' midyear. I dropped `3.10` because it dies in fall, but maybe it's too premature to drop `3.11`.

membrane_curvature/fft_filtering.py
ojeda-e: ooof :( yeah, I added a check for this and same for below.

docs/source/pages/Usage.rst
ojeda-e: ok, fixed now 
image

ojeda-e:
Thank you so much for the review @tylerjereddy! I addressed most of your comments -didn't modify the workflow nit you mentioned but I'll keep that on my radar.

Edit: I opened an issue to add tests to cov report. There are some lines I need to tweak a little bit because the coverage is going down to 98%. I'll address that separetely. Thanks again!

PR #193, opened 2026-06-04
Docs: enhance RTD navigation menu for API docs

### LINKED ISSUE #189: Docs: Navigation panel for API docs is too crowded
After adding new features (#146 and #188) the navigation menu on the left of the documentation looks too crowded. As it is today, it's difficult to navigate and makes finding API docs a bit unclear. Specially because there are several private functions.

Not sure what's the best approach, but there must be a better way to display the menu, with collapsable sections (in green), that makes navigating the API docs menu (in red) more intuitive and easier. Ideally this change should be shipped in v2.0.0

Image

ojeda-e:
Fixes #189 

## Description

Set sections in navigation menu for API docs.

Refactored API stubs: 
 - api now includes surface_methods and fourier_validators
   - surface_methods: groups binning and fourier.
   - validators now include fourier_validators (this is is useful because in the future there may be more).
   - curvature now groups all curvature functions.

## PR Checklist

 - [N/A] Tests updated/added?


docs/api/binning_surface.rst
```diff
@@ -1,4 +1,9 @@
+Binning Surface
```
ojeda-e: Moved the titles to the stubs. Not sure about this change, but the tree looks better with this change:

Image

docs/api/fourier_validators.rst
```diff
@@ -1,5 +1,10 @@
...
+==================
+
+.. toctree::
+   :maxdepth: 1
+
 
 .. automodule:: membrane_curvature.fourier_validators
-    :members:
-    :private-members: _coerce_positions
-    :member-order: bysource
```
ojeda-e, line 5: remove the order bysource. Now the public functions show first, and then the private ones.

membrane_curvature/fourier_surface.py
```diff
@@ -1,9 +1,5 @@
 r"""
 
----------------
```
ojeda-e, line 3: This is a side by side comparion of how it looks moving this title to the .rst file vs leaving it here.

Image

docs/api/fourier_surface.rst
```diff
@@ -1,5 +1,9 @@
+Fourier Surface
```
ojeda-e: Moved the titles to the stubs. Not sure about this change, but the tree looks better with this change:


```
Surface Methods
  -> Fourier surface
       --> Functions
       --> f1
       --> f2 (etc)
```
Image

ojeda-e (commented):
The changes achieve the main goal, which is making the navigation menu more readable and organized by having collapsable sections.

Image

However, I am not sure about moving the title to the .rst, and this is the best solution I found. Open to suggestions if there is a better way.

ojeda-e:
@lilyminium may I ask you for some suggestions here if you have some spare time? If you are available please feel free to leave comments or request changes and I'll add your tag as a reviewer. Thanks!

docs/api/fourier_surface.rst
ojeda-e: @orbeckst do you have any opinions or suggestions here? I know MDAnalysis doesn't set titles in the `.rst` files, but ignoring the reasons behind it. My reasons are purely aesthetic in this case.

orbeckst: Do whatever works best for you. Docs are often a compromise between writing any docs at all (typically inside the code and then using autosphinx) vs "perfectly" crafted explicit rst.

ojeda-e: Thanks, I revised the changes and found out the best version (IMO) can be done with more tweaks w/o removing this titles, I just needed more time to play and move. The last push has all the changes and I really like it!

membrane_curvature/base.py
```diff
@@ -111,39 +128,6 @@ class MembraneCurvature(AnalysisBase):
     :class:`~MDAnalysis.transformations.wrap.wrap`
         Wrap/unwrap the atoms of a given AtomGroup in the unit cell.
 
-    Notes
```
ojeda-e, line 114: this was duplicated, I removed one of them.

ojeda-e:
After several iterations from the initial changes, I finally have the layout I was looking for while achieving the main goal of this PR: fixing the navigation menu. 
I love the way it looks with the changes in this PR and I'm hppy to merge.

Here a snapshot of the entrypoint page for API docs for a reference:

image

Issue #143, opened 2026-04-25
Verify install steps in `deploy` workflow are flaky because of PyPI index propagation takes too long


ojeda-e:
This is not urgent but should def be checked before next release.

After publishing [v1.1.2](https://pypi.org/project/membrane-curvature/1.1.2/), the workflow’s [Verify install from PyPI (macos-latest)](https://github.com/MDAnalysis/membrane-curvature/actions/runs/24918302627/job/72975632135#logs) and [Verify install from PyPI (ubuntu-latest)](https://github.com/MDAnalysis/membrane-curvature/actions/runs/24918302627/job/72975632358#logs) failed on the first run because pip could not resolve the latest package version after upload. This is likely happening because the CDN/index takes a long time to fully propagate. I found a similar behaviour for TestPyPI.

At the moment, the deploy workflow uses a fixed sleep 30 before installing the just-published version from the index:

https://github.com/MDAnalysis/membrane-curvature/actions/runs/24918302627/workflow#L157

Seems like this strategy doesn't give enough time to catch up, so verification is flaky and I needed manual re-runs.

A better approach is to have a retry with timeout until the package is available and we can validate installation.

ojeda-e:
I did a pre-release today and I didn't find the behaviour initially described. Regardless, I will keep this issue open to track whenever it happens again. 
I am assigning a low priority to this issue and removing it from the v2.0.0 milestone.

ojeda-e:
While working on #194 , the error showed up again in:
- TestPyPI
https://github.com/MDAnalysis/membrane-curvature/actions/runs/27049549511/job/79842401270#step:5:28
(This time in ubuntu-latest only)

- PyPi, ubuntu-latest and macos-latest
https://github.com/MDAnalysis/membrane-curvature/actions/runs/27050370435/job/79844918040#step:5:26
https://github.com/MDAnalysis/membrane-curvature/actions/runs/27050370435/job/79845023557#step:5:26

orbeckst:
Feel free to use the workflow https://github.com/Becksteinlab/multibind/blob/master/.github/workflows/deploy.yaml with the custom wait action that waits until the package is up. The code is MIT so should be useable.

(In other cases I just added an unsophisticated sleep 30.)

ojeda-e:
[...]

PR #194, opened 2026-06-06
Modernize: Comply with PEP 735

### LINKED ISSUE #191: Comply with PEP 735
To fix this issue, modernize `pyproject.toml` to comply with [PEP 735](https://peps.python.org/pep-0735/).

Thanks for the suggestion @tylerjereddy

FWIW, conformance with the PEP has been tricky for other projects--SciPy has been stuck for a bit with 639 here: https://github.com/scipy/scipy/pull/23451

If you're modernizing `pyproject.toml` there's also the `project.optional-dependencies` to `[dependency-groups]` migration (PEP 735) where appropriate. We haven't finished that yet upstream either, https://github.com/scipy/scipy/issues/25050.

_Originally posted by @tylerjereddy in https://github.com/MDAnalysis/membrane-curvature/pull/185#discussion_r3344859947_

ojeda-e:
Fixes #191 

## Description

- Switch optional deps for group deps in pyproject.toml
- Update installation instructions in README and Getting Started 
- In CI, install tests w/ --groups 

## PR Checklist

 - [N/A] Tests updated/added?

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -154,6 +154,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          path: repo
 
       - name: Set up Python
         uses: actions/setup-python@v6
@@ -167,7 +169,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ "membrane_curvature==${{ needs.build.outputs.version }}"
-          pip install --group tests
+          pip install --group repo/pyproject.toml:tests
 
       - name: Test import
         run: |
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -154,8 +154,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
-        with:
-          path: repo
 
       - name: Set up Python
         uses: actions/setup-python@v6
@@ -169,7 +167,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ "membrane_curvature==${{ needs.build.outputs.version }}"
-          pip install --group repo/pyproject.toml:tests
+          pip install --group tests
 
       - name: Test import
         run: |
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -64,6 +64,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: build
     steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
@@ -79,7 +82,8 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           WHEEL_FILE=$(ls dist/*.whl)
-          pip install "$WHEEL_FILE"[tests]
+          pip install "$WHEEL_FILE"
+          pip install --group tests
 
       - name: Test import
         run: |
@@ -159,7 +163,8 @@ jobs:
       - name: Install from TestPyPI
         run: |
           python -m pip install --upgrade pip
-          pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ "membrane_curvature[tests]==${{ needs.build.outputs.version }}"
+          pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ "membrane_curvature==${{ needs.build.outputs.version }}"
+          pip install --group tests
 
       - name: Test import
         run: |
@@ -195,7 +200,8 @@ jobs:
       - name: Install from PyPI
         run: |
           python -m pip install --upgrade pip
-          pip install "membrane_curvature[tests]==${{ needs.build.outputs.version }}"
+          pip install "membrane_curvature==${{ needs.build.outputs.version }}"
+          pip install --group tests
 
       - name: Test import
         run: |
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -152,6 +152,9 @@ jobs:
       github.repository == 'MDAnalysis/membrane-curvature' &&
       (github.event_name == 'push' && startsWith(github.ref, 'refs/tags/'))
     steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
--- a/README.md
+++ b/README.md
@@ -171,16 +171,23 @@ available where you can find:
 
 ### Contributing
 
-Contributions are very welcome!
+Contributions are very welcome! Check the [Contributing] page for more details.
+
+If you are interested in contributing to MembraneCurvature, installation of the development dependencies
+is required. There are three dependency groups defined in `pyproject.toml`:
+
+- `tests`: testing dependencies.
+- `docs`: documentation build dependencies (e.g. Sphinx, themes).
+- `dev`: development tools (includes `tests` and `docs`).
+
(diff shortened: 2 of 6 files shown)


ojeda-e:
CI passes so far.
Installation with the former optional dependencies are in the `deploy.yaml` workflow. I'll test releasing from a branch to check everything works.

ojeda-e:
Deployment [failed](https://github.com/MDAnalysis/membrane-curvature/actions/runs/27049399981/job/79841981974) in Verify install from TestPyPI.  Testing adding a checkout,

ojeda-e:
Okay, deployment passes for [TestPyPI](https://github.com/MDAnalysis/membrane-curvature/actions/runs/27050060674).

ojeda-e:
And deployment to [PyPI passes](https://github.com/MDAnalysis/membrane-curvature/actions/runs/27050370435)! 🎉 
I had to re-run the worflows, tho. Added in comment in #143 to keep tracking this behaviour.

pyproject.toml
```diff
@@ -65,7 +65,9 @@ dev = [
     "ruff",
     "validate-pyproject[all]",
     "pre-commit",
-    "ty"
+    "ty",
+    {include-group = "tests"},
```
ojeda-e, line 68: The `include-group` are the only thing I'm not sure about. My rationale is: if someone wants to install all dev deps, why would I need three different groups to do so? 
- With `--groups dev` everything is included.
- If the user wants `--groups tests` or `--group docs` they're also available individually.

`dev` includes everything.
 I like the style, but maybe this is not a good practice?

ojeda-e:
Thanks for suggesting this addition @tylerjereddy. I'm adding you as a reviewer for whenever you have time. Thanks!

README.md
```diff
@@ -232,3 +243,4 @@ project with [MDAnalysis] and it is linked to a [Code of Conduct][code_of_conduc
 [pre-commit]: https://pre-commit.com
 [ruff]: https://docs.astral.sh/ruff/
 [uv]: https://docs.astral.sh/uv/
+[Contributing]: https://github.com/MDAnalysis/membrane-curvature/blob/main/.github/CONTRIBUTING.md
```
ojeda-e, line 248: @orbeckst I am pointing to [this](https://github.com/MDAnalysis/membrane-curvature/blob/main/.github/CONTRIBUTING.md) Contributing page. Not sure MDAkits have a standard contributing guide, or if because the repo belongs to the MDAnalysis organization this should be more consistent with other guidelines.

orbeckst: That seems perfect. Kits should defined themselves how contributions are handled.

.github/workflows/deploy.yaml
```diff
@@ -64,6 +64,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: build
     steps:
+      - name: Checkout
+        uses: actions/checkout@v6
```
tylerjereddy, line 68: Probably the last thing you want is more work -- maybe for another time/issue -- it is a good practice to use hashes rather than the version numbers/tags for security purposes. I think that's because it is much harder to compromise a hash than a release tag. But that's for another time...

(if you do that across the codebase, you can set i.e., dependabot to update the hashes for you with new versions, typically with 1-2 week cooldown for supply chain security)

It would typically look something like:

`- uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2`

then `dependabot` would update the hash and tag/version automatically as needed

docs/getting_started.rst
```diff
@@ -29,42 +29,53 @@ The following command will install or upgrade the latest stable version of Membr
...
 
 .. code-block:: bash
 
-   pip install -e ".[dev,tests,docs]"
+   pip install -e .
+   pip install --group dev
+
+
+Note that MembraneCurvature defines development dependency groups in ``pyproject.toml``, which are 
+not included in the published package, so installation of these dependencies is required.
```
tylerjereddy: required for development but not end user runtime, right?

docs/getting_started.rst
```diff
@@ -29,42 +29,53 @@ The following command will install or upgrade the latest stable version of Membr
...
 
-Installation from source is also available with pip by cloning the repository and running:
+Installation from source is also available with pip by cloning the repository and running the following commands:
 
 .. code-block:: bash
 
    git clone https://github.com/MDAnalysis/membrane-curvature.git
    cd membrane-curvature
    python -m pip install -e .
+   pip install --group dev
```
tylerjereddy, line 79: same here--only needed if a user wants to *develop* your package (or test it), right?

docs/getting_started.rst
```diff
@@ -102,18 +113,12 @@ To install the latest stable version of MembraneCurvature with `uv`_, run the fo
 
    uv pip install membrane-curvature
 
-For development, you can install the optional dependencies with the following commands:
+For development, install the development dependency group with the following commands:
```
tylerjereddy: I guess I'm maybe slightly confused about using the dev dep group above if this is the actual development section of the install docs.

pyproject.toml
tylerjereddy: you need `ipython` to run the testsuite?

tylerjereddy, line 50: similar question here

pyproject.toml
tylerjereddy: I think you're correct here yep!

If your project has genuine optional runtime dependencies that a user *might* want to use, those are still perfectly fine (and good) to retain in a separate pared down `optional-dependencies` (i.e., no need to delete that section if you still genuinely need it). Do you have any?

tylerjereddy (commented):
I took a look at this one first because the diff is smaller. Looks pretty solid overall. You're definitely doing this before most other projects do!

.github/workflows/deploy.yaml
ojeda-e: Thanks for the suggestion, def a good idea. Will open an issue.

docs/getting_started.rst
ojeda-e: yes, correct. Maybe I should clairfy that.

docs/getting_started.rst
ojeda-e: damn, this isn't supposed to be here. This is to install from source, but not installing dev. Will fix this.

docs/getting_started.rst
ojeda-e: Thanks for catching this, yes, you are correct. That should be installation from source but w/o dev deps.

pyproject.toml
ojeda-e: I'm kind of surprised to see this here. Will double check.
Edit: No, this is not supposed to be here. Thanks!

ojeda-e: This one in docs is needed for sure. I have some tutorials in notebooks and Sphinx needs it to render the notebook properly.

pyproject.toml
ojeda-e: Thanks for confirming.
I have no optional deps, really. Everything is covered with MDAnalysis and Numpy.

ojeda-e (commented):
Thanks for your comments! Will update the getting started in the next push.

docs/getting_started.rst
ojeda-e: it looks odd in this diff, but I updated the doc. It should look better now.

ojeda-e:
The CI was failing in the last commits pushed today. Happens that 2 days ago the codecov action had an upgrade. I fixed it by bumping  to v7.
See https://github.com/codecov/codecov-action/issues/1956 for more details.

docs/getting_started.rst
ojeda-e: added a clarifying note to reinforce the text.

tylerjereddy:
I was happy enough with the diff, and CI was passing so I merged it.

I think there was maybe one grammar thing I didn't bother commenting on, and I suppose you could think about splitting out "linting"/"typing" deps, but that's kind of a separate matter vs. the PEP conformance anyway.

Re: "uv" as the preferred tool--if `pixi` isn't on your radar yet, it is gaining a lot of traction in the ecosystem too--pretty great for small research projects too b/c of the dep locking (so if you're working with a student or something you could basically guarantee locking of the same environment).

ojeda-e:
Thank you! I'll check on `pixi`.

PR #200, opened 2026-06-13
Add initial ASV benchmarks

### LINKED ISSUE #84: Benchmark performance
Currently, there is no information about the performance of membrane curvature. It would be nice to have a page in the documentation that addresses this matter. Although the idea is not very well defined yet, some guidelines for an initial version of the benchmark are:

- Performance over number of lipids in single-frame systems. 
- Performance over number of frames with a fixed number of lipids. 
- Keep the list of assumptions regardless of the approach.
- Consider dummy Universes. 
- MDAnalysis uses ASV (Check code [here](https://github.com/MDAnalysis/mdanalysis/tree/develop/benchmarks))

Thanks @hmacdope, @lilyminium and @richardjgowers for the suggestions.

ojeda-e:
Fixes #84 

## Description

- Added ASV json config with uv envs.
- Included outputs of benchmarks in `.gitignore`.
- Added basic benchmark.
- Created README with steps and brief explanations to run ASV.

## PR Checklist

 - [N/A] Tests updated/added?

**Notes:**

- Right now everything lives in a single file (`benchmarks/membranecurvature.py`), to cover `MembraneCurvature.run()` performance across system sizes, surface methods (Fourier and binning), Fourier modes, and multi-frame trajectories. 
- I’m treating this as a first pass rather than the final benchmark layout. Longer term, I’d like to split benchmarks into one file per module under `membrane_curvature/ ` For example `base`, `binning_surface`, `fourier_surface`, etc. Then we could track performance at both the analysis API level and for lower-level building blocks. This is particularly interesting because there is a difference in performance between binning and fourier, but also one when increasing the fourier modes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,3 +112,10 @@ ENV/
 
 # Generated docs figs
 docs/savefig/
+
+# ASV benchmarks
+benchmarks/results/
+benchmarks/html/
+benchmarks/env/
+benchmarks/virtualenv/
+benchmarks/.asv/
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -1,0 +1,187 @@
+MembraneCurvature benchmarks
+=============================
+
+Performance benchmarks for MembraneCurvature using [Airspeed Velocity] (ASV).
+
+ASV creates isolated environments with the [uv plugin] (See [environment_type: uv] in [asv.conf.json]).
+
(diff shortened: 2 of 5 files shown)


ojeda-e:
@orbeckst I started with the approach you suggested (minimal/simple becnhmarks) with its config and README. For anytime you are available I'll be happy to receive your feedback. Thanks!

PR #202, opened 2026-06-17
Build: Add uv lock file and verify dev group installation 

### LINKED ISSUE #201: Missing uv lock file with unclear dev installation instructions.
Instructions in docs and README say that `uv` is the recommended tool for dev, but there is no `uv.lock` file in the project. There are also some inconsistencies in the doc pages that is best to fix to avoid conflicting information.

To fix this issue:
- [x] Add `uv.lock` file.
- [x] Verify install from a git clone with `uv sync --frozen`.
- [x] Align README and `docs/getting_started.rst` dev instructions.
- [x] Add uv lock notes to `.github/CONTRIBUTING.md`

## CI (optional)
Tests in CI currently run with conda, which matches what most MDAkits do and what I expect from the broader MDAnalysis ecosystem. I still think we should guard the documented `uv` dev path in CI somehow. Otherwise if something goes wrong with `uv.lock` or `uv sync`, it'll only get noticed when a contributor clones the repo and hits the error.

I'm tempted to add a minimal CI job that:
- [x] Runs  `uv sync --frozen`.
- [x] Run tests with uv, min Python version (3.11), and single platform (Linux or Mac).

With this change, my intention is to catch a "bad" or stale `uv.lock` file when `pyproject.toml` keys change. That is probably the valuable part of this adition. Then, if we add running tests with uv, we would also validate the exact environment contributors are told to use. For this I'm suggesting one single python version rather than a matrix, and a one platform only to avoid redundancy.

Either way, I would keep the existing conda test matrix as it is now rather than replacing it. Recommending `uv` for dev doesn't have to mean moving the whole project off conda for the CI runs.

@tylerjereddy @RMeli @orbeckst If you have any thoughts/opinions on the optional CI job, I'd love to hear them. I'm tempted to add this, but better to invest time on something that is worth adding.

ojeda-e:
Fixes #201 

## Description

After adding `uv.lock`, verified uv installation for dev deps is functional. I cloned the repo and ran:
- `sync --frozen --group dev`.
- `uv run pytest` - 
- `uv run pre-commit run --all-files`.
- `cd docs && uv run make html`.
 
all of them worked as expected.

- [x] Added `uv.lock`.
- [x] Updated CONTRIBUTING guide to add uv dev env instructions.
- [x] Aligned dev instructions in getting started and README.
- [x] Added step in CI to run tests in uv dev env  - py311 only.

## PR Checklist

 - [N/A] Tests updated/added?

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -71,7 +71,9 @@ For more information on building the documentation, see the [docs/README.md](../
   and multiple core developers give "Approved" reviews.
 
 > [!NOTE]
-> All tests must pass before a PR can be merged. CI runs the test suite, linters, and type checks. Coverage is reported to Codecov and it is reported in the PR. Maintainers may block merge if code coverage drops below 100%.
+> All tests must pass before a PR can be merged. CI runs the test suite, linters, and type checks.
+> New code warrants new tests, and all added code should be covered by the test suite.
+> Coverage is reported to Codecov and it is reported in the PR. Maintainers may block merge if code coverage drops below 100%.
 
 # Additional Resources
 
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -11,6 +11,41 @@ describes how to merge code changes into this membrane_curvature.
   [clone](https://help.github.com/articles/cloning-a-repository/) your fork of
   the repository.
 
+## Development Environment
+
+We strongly recommend using [uv](https://docs.astral.sh/uv/) as a development environment for MembraneCurvature. To set up a `uv` dev environment, run the following commands from the root of the repository:
+
+```bash
+uv sync --group dev
+uv run pre-commit install
+```
+
+### Running tests
+
+Tests are run with [pytest](https://docs.pytest.org/en/stable/) and are a fundamental part of the development process. All tests must pass before a PR can be merged.
+
+To run the tests:
+
+```bash
+uv run pytest
+```
+
+To get the coverage report, run the tests with the `--cov` flag:
+
+```bash
+uv run pytest membrane_curvature/tests  --cov=membrane_curvature --cov-report=html --cov-report=term
+```
+
+### Building the documentation
+
+To build the documentation, run the following command from the root of the repository:
+
+```bash
+uv run sphinx-build -b html docs/ build/html
+```
+
+For more information on building the documentation, see the [docs/README.md](../docs/README.md) file.
+
 ## Making Changes
 
 * Add some really awesome code to your local fork.  It's usually a [good
@@ -18,7 +53,8 @@ describes how to merge code changes into this membrane_curvature.
   to make changes on a
   [branch](https://help.github.com/articles/creating-and-deleting-branches-within-your-repository/)
   with the branch name relating to the feature you are going to add.
-* When you are ready for others to examine and comment on your new feature,
+* Add the relevant changes to the [CHANGELOG.rst](../CHANGELOG.rst) file and add your name to the [AUTHORS](../AUTHORS) file if it's your first contribution.
+* When you are ready for others to review and comment on your new feature,
   navigate to your fork of membrane_curvature on GitHub and open a [pull
   request](https://help.github.com/articles/using-pull-requests/) (PR). Note that
   after you launch a PR from one of your fork's branches, all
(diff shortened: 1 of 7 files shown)


ojeda-e:
I played with the CI to have an idea of how uv runs tests vs micromamba (see [usage](https://github.com/MDAnalysis/membrane-curvature/actions/runs/27708831960/usage)). It seems like uv runs much faster, but that discussion is outside the scope of this PR. I'll open an issue to assess this separetely.

For now I'm adding a step to the CI to run tests with the uv env that installs the dev dependency group.

.github/workflows/CI.yaml
```diff
@@ -58,6 +58,20 @@ jobs:
         - name: Run ty - check
           run: ty check
 
+  uv-dev-check:
+    name: Validate uv environment
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - name: Install uv and set Python 3.11
+      uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b
```
ojeda-e, line 68: Adding the action with the commit hash instead :)

.github/workflows/CI.yaml
```diff
@@ -58,6 +58,20 @@ jobs:
...
+      with:
+        python-version: "3.11"
+    - name: Create uv env with dev dependency group
+      run: uv sync --frozen --group dev
```
ojeda-e, line 72: I'm running with frozen because we are installing exactly what's in the `uv.lock` file, so it fails if `uv.lock` and `pyproject.toml` don't match.

README.md
```diff
@@ -169,54 +169,51 @@ available where you can find:
...
-
-To set up hooks locally, with [uv]:
+To create a new `uv` environment and install the dependencies included in the `dev` group, run:
 
 ```bash
 uv sync --group dev
-uv run pre-commit install
 ```
 
-Or with pip :
```
ojeda-e, line 206: Intentionally remove this part to convey in a clearer manner than `uv` is preferred. With this I intend to focus on `uv`, and redirect the user to another page if their choice is pip.

README.md
```diff
@@ -169,54 +169,51 @@ available where you can find:
...
+To run the pre-commit hooks manually, with [uv]:
 
 ```bash
-pip install -e .
-pip install --group dev
-pre-commit install
+uv run pre-commit run --all-files
 ```
 
+For instructions on how to create a development environment with pip, see the [Installing development dependencies with pip] page.
```
ojeda-e, line 210: Here I redirect the user to installation of dev env with pip.

.github/workflows/CI.yaml
```diff
@@ -58,6 +58,20 @@ jobs:
         - name: Run ty - check
           run: ty check
 
+  uv-dev-check:
+    name: Validate uv environment
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - name: Install uv and set Python 3.11
```
RMeli, line 66:
```suggestion
    - name: Install uv and set Python
```
Nit: just because this might go out of sync with the actual version of Python set below.

.github/workflows/CI.yaml
RMeli: Why a specific commit is needed, instead of a version pin? Maybe add a comment on why this is neede?

.github/CONTRIBUTING.md
```diff
@@ -34,6 +70,9 @@ describes how to merge code changes into this membrane_curvature.
   integration returns checkmarks,
   and multiple core developers give "Approved" reviews.
 
+> [!NOTE]
+> All tests must pass before a PR can be merged. CI runs the test suite, linters, and type checks. Coverage is reported to Codecov and it is reported in the PR. Maintainers may block merge if code coverage drops below 100%.
```
RMeli:
```suggestion
> All tests must pass before a PR can be merged. CI runs the test suite, linters, and type checks.
> New code warrants new tests, and all added code should be covered by the test suite.
> Coverage is reported to Codecov and it is reported in the PR. Maintainers may block merge if code coverage drops below 100%.
```

RMeli (approved):

.github/workflows/CI.yaml
ojeda-e: In general using actions with hashes is preferred for reliability/security, because it is unique and inmmutable. In this specific case, it is also because this is the [usage](https://github.com/astral-sh/setup-uv#usage) suggested in docs for this action, which is equivalent to v8.1.0.
I'll add a comment to clarify that.

.github/workflows/CI.yaml
ojeda-e: This one I'll skip because I feel the title is more informative with the python version and will save me time in the future when reading from the actions tab.

ojeda-e:
Thanks for your questions and suggestions @RMeli!

PR #221, opened 2026-07-09
Refact: Remove side effect - MDAnalysis logging at import

### LINKED ISSUE #207: Remove MDAnalysis.start_logging() side effect from module import
While running some analyses with `v2.0.0rc6` I noticed duplicated logs. I think this happens because  `MDAnalysis.start_logging()` is called at import time in:

https://github.com/MDAnalysis/membrane-curvature/blob/db30081195c26238dd5472b298b42a580cd55807/membrane_curvature/base.py#L51

and

https://github.com/MDAnalysis/membrane-curvature/blob/db30081195c26238dd5472b298b42a580cd55807/membrane_curvature/binning_surface.py#L35

That configures logging (console + MDAnalysis.log) whenever the package is imported.

MDAnalysis documents this for application code, not libraries. The important user messages already use `warnings.warn()`, so removing these calls should not change typical behavior.

Remove both `start_logging()` calls and keep module loggers. Document that users can call MDAnalysis.start_logging() in their own scripts if they want file logging.

ojeda-e:
Fixes #207 

## Description

- Removed `MDAnalysis.start_logging()` from `base.py` and `binning_surface.py` (kept module loggers `logger = logging.getLogger`).
- Added basic usage example w/ and w/o logger in getting started.

## PR Checklist

 - [N/A] Tests updated/added?

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -153,6 +153,73 @@ dependency groups (including ``tests`` and ``docs``) in one step.
 
 Installation guides :ref:`with_pip` and :ref:`with_conda` are also available if you prefer them.
 
+Basic example
+-------------
+
+This is an example on how to use MembraneCurvature:
+
+.. code-block:: python
+
+      import MDAnalysis as mda
+      from membrane_curvature.base import MembraneCurvature
+      from MDAnalysis.tests.datafiles import XTC_MEMPROT, GRO_MEMPROT  # test trajectory
+
+      # create a universe from the coordinates and trajectory files
+      universe = mda.Universe(GRO_MEMPROT, XTC_MEMPROT)
+
+      # run the membrane curvature analysis
+      mc = MembraneCurvature(universe,
+                             select='resid 297-517 and name P'
+                             ).run()
+
+      # access the results
+      surface = mc.results.average_z_surface
+      mean_curvature = mc.results.average_mean
+      gaussian_curvature = mc.results.average_gaussian
+
+You can find more details on how to use MembraneCurvature in the `Usage`_ page.
+
+Please note that **MembraneCurvature does not configure MDAnalysis logging on import**.
+If you want MDAnalysis log output, set up your own logger as described in the `MDAnalysis logging guide`_. 
+A basic example of how to set up your own logger is shown below.
+
+.. code-block:: python
+
+      import logging
+      import MDAnalysis as mda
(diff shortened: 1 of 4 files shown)


membrane_curvature/base.py
```diff
@@ -43,13 +43,10 @@
 from .fourier_surface import n_fourier_parameters
 from .fourier_validators import validate_positive_bin_counts
 
-import MDAnalysis
```
ojeda-e, line 46: @orbeckst do you have any objections? I'd like to have your input here.

orbeckst: You don't have to use the MDAnalysis start_logging() function — that's just a convenience. Totally fine to remove it here.

I think @jauy123 (eg https://github.com/MDAnalysis/mdanalysis/issues/5357 ) also has some thoughts on how to improve the default logging facility.

orbeckst: In fact, definitely remove it. It makes much more sense to set up the logger in a top-level script.

membrane_curvature/binning_surface.py
```diff
@@ -29,10 +29,8 @@
 
 import numpy as np
 import warnings
-import MDAnalysis
 import logging
 
-MDAnalysis.start_logging()
```
orbeckst, line 35: Definitely remove it. It makes much more sense to set up the logger in a top-level script.

docs/getting_started.rst
```diff
@@ -153,11 +153,80 @@ dependency groups (including ``tests`` and ``docs``) in one step.
...
+.. code-block:: python
+
+      import logging
+      import MDAnalysis as mda
+      from membrane_curvature.base import MembraneCurvature
+      from MDAnalysis.tests.datafiles import XTC_MEMPROT, GRO_MEMPROT  # test trajectory
+
+      # set up the logger
+      logging.basicConfig(level=logging.INFO)
+      logger = logging.getLogger("MDAnalysis.MDAKit.membrane_curvature")
```
orbeckst: This is where you _could_ use `mda.start_logging()` (but you really don't have to).

orbeckst (approved):
Good idea to remove the start_logging() from the modules.

docs/getting_started.rst
ojeda-e: Agreed, thanks for catching this. Updated.

PR #223, opened 2026-07-09
Refact: Calculate Fourier coeffs in SVD with `np.dot` instead of matmul

### LINKED ISSUE #222: Bug: Warnings in Fourier coefficients are spurious Numpy messages
While working on #221 and running the example provided in the Getting Started doc page:

```python
import MDAnalysis as mda
from membrane_curvature.base import MembraneCurvature
from MDAnalysis.tests.datafiles import XTC_MEMPROT, GRO_MEMPROT  # test trajectory

# create a universe from the coordinates and trajectory files
universe = mda.Universe(GRO_MEMPROT, XTC_MEMPROT)

# run the membrane curvature analysis
mc = MembraneCurvature(universe,
                       select='resid 297-517 and name P'
                       ).run()

# access the results
surface = mc.results.average_z_surface
mean_curvature = mc.results.average_mean
gaussian_curvature = mc.results.average_gaussian

```
I got the output:

```text
membrane_curvature/fourier_surface.py:370: RuntimeWarning: divide by zero encountered in matmul
  coefficients_on_span = (unitary_array[:, keep].T @ targets) / singular_values[keep]
membrane_curvature/fourier_surface.py:370: RuntimeWarning: overflow encountered in matmul
  coefficients_on_span = (unitary_array[:, keep].T @ targets) / singular_values[keep]
membrane_curvature/fourier_surface.py:370: RuntimeWarning: invalid value encountered in matmul
  coefficients_on_span = (unitary_array[:, keep].T @ targets) / singular_values[keep]
```

The warnings come from Numpy because combining `@` with `/` on the same line can emit invalid value warnings even when the result is correct. The solution is probably just splitting the line for `coefficients_on_span` into 2 separate calculations.

ojeda-e:
Fixes #222

## Description

- Split `coefficients_on_span` into `projection` (with np.dpt) and `singular_values`.
- Updated tests. Updated fixture with larger size and random values for design matrix and targets. 

## PR Checklist

 - [x] Tests updated/added?

### DIFF
--- a/membrane_curvature/tests/test_fourier_surface.py
+++ b/membrane_curvature/tests/test_fourier_surface.py
@@ -2,6 +2,7 @@
 Unit tests for Fourier surface functions.
 """
 
+import warnings
 from dataclasses import dataclass
 
 import numpy as np
@@ -105,17 +106,12 @@ def dummy_two_mode_fourier_system(dummy_fourier_universe):
 
 @pytest.fixture()
 def full_rank_design_system():
-    """Full-rank design with a known exact least-squares solution."""
-    design_matrix = np.array(
-        [
-            [1.0, 0.0],
-            [0.0, 1.0],
-            [1.0, 1.0],
-        ],
-        dtype=np.float64,
-    )
-    theta_true = np.array([2.0, 3.0], dtype=np.float64)
-    targets = design_matrix @ theta_true
+    """Full-rank design with a known exact least-squares solution. Random values for design matrix and targets."""
+    rng = np.random.default_rng(0)
+    n_rows, n_columns = 100, 25
+    design_matrix = rng.standard_normal((n_rows, n_columns))
+    theta_true = rng.standard_normal(n_columns)
+    targets = np.dot(design_matrix, theta_true)
     return design_matrix, targets, theta_true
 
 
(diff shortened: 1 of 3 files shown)


membrane_curvature/tests/test_fourier_surface.py
```diff
@@ -105,17 +106,12 @@ def dummy_two_mode_fourier_system(dummy_fourier_universe):
 
 @pytest.fixture()
 def full_rank_design_system():
-    """Full-rank design with a known exact least-squares solution."""
-    design_matrix = np.array(
```
ojeda-e, line 109: This previous `full_rank_design_system` was 3×2, which doesn't trigger the `@` + `/` spurious warning. This only shows up once the design matrix is large enough.
Edit: Not sure if I should keep the small one for some of the other tests (this fixture is used in 5 unit tests). And then use this large one to check the warning is not triggered. I personally prefer having the same fixture for all the tests

ojeda-e:
@tylerjereddy in case you have some time to check it out, this is a very small change :)

I basically had two options to update the tests:

1) Update the fixture, make it larger and use it to check the spurious warning wasn't triggered.
2) Keep using the small fixture and create a large dummy example to test the warning is not triggered anymore.

I opted for 1, but please correct me if there is a better way than the two listed here. Thanks!

PR #224, opened 2026-07-09
CI: Replace action versions w/ commit SHA + dependabot cfg file

### LINKED ISSUE #196: CI: Update actions to use hashes rather than version numbers
I was hesitating having DependaBot in MembraneCurvature, I thought it was overkilling for the project since there are only 2 main dependencies. However, this could be a very good use case for it.

To fix this issue:
- [x] Replace version numbers/tags with hashes.
- [x] Add dependabot. 

Probably the last thing you want is more work -- maybe for another time/issue -- it is a good practice to use hashes rather than the version numbers/tags for security purposes. I think that's because it is much harder to compromise a hash than a release tag. But that's for another time...

(if you do that across the codebase, you can set i.e., dependabot to update the hashes for you with new versions, typically with 1-2 week cooldown for supply chain security)

It would typically look something like:

`- uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2`

then `dependabot` would update the hash and tag/version automatically as needed

_Originally posted by @tylerjereddy in https://github.com/MDAnalysis/membrane-curvature/pull/194#discussion_r3370438463_

### LINKED ISSUE #219: Add dependabot.yaml file
Context: Dependabot has been enabled in all the repos org-wide. 

To fix this issue:

- [x] Add a `dependabot.yml` file 

It can be minimal to start. See https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/auto-update-actions

ojeda-e:
Fixes #196 and #219 

Dependabot has been enabled for all repos in the organization.

## Description

Replaced:
- checkout action version with hash v7.0.0
- setup python action version with hash v6.2.0
- download artifact action version with hash v8.0.1
- upload artifact action version with hash v7.0.1
- publish PyPI v1.14.0
- setup micromamba action version with hash v1.0.0
- codecov action version with hash v7.0.0
- Added dependabot config file

I upgraded some actions in the process too:
- checkout v6 -> v7
- setup-python v6 -> v6.2.0

## PR Checklist

 - [N/A] Tests updated/added?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,37 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "sunday"
+      time: "18:00"
+    groups:
+      all-actions:
+        patterns: [ "*" ]
+
+  - package-ecosystem: "uv"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "saturday"
+      time: "18:00"
+    groups:
+      production:
+        patterns: ["*"]
+        dependency-type: "production"
+      development:
+        patterns: ["*"]
+        dependency-type: "development"
+      ignore:
+      - dependency-name: "nglview"
+
+  - package-ecosystem: "conda"
+    directory: "/devtools/conda-envs"
+    schedule:
+      interval: "weekly"
+      day: "saturday"
+      time: "18:00"
+    groups:
+      all-conda:
+        patterns: ["*"]
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,7 +23,7 @@ updates:
       development:
         patterns: ["*"]
         dependency-type: "development"
-      ignore:
+    ignore:
       - dependency-name: "nglview"
 
   - package-ecosystem: "conda"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,8 @@ updates:
       interval: "weekly"
       day: "sunday"
       time: "18:00"
+    cooldown:
+      default-days: 7
     groups:
       all-actions:
         patterns: [ "*" ]
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
-    - uses: actions/setup-python@v6
+    - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
       with:
         python-version: "3.11"
     - name: Install pre-commit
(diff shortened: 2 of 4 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,37 @@
...
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "sunday"
+      time: "18:00"
+    groups:
+      all-actions:
+        patterns: [ "*" ]
```
ojeda-e, line 13: @orbeckst I added the config file for dependabot. The first part is to update the actions,  and then it's for uv and conda deps.

.github/dependabot.yml
```diff
@@ -0,0 +1,37 @@
...
+      time: "18:00"
+    groups:
+      production:
+        patterns: ["*"]
+        dependency-type: "production"
+      development:
+        patterns: ["*"]
+        dependency-type: "development"
+    ignore:
+      - dependency-name: "nglview"
```
ojeda-e, line 29: Ignoring nglview because currently the docs only build with 4.0.0 only despite of being higher versions available.

.github/dependabot.yml
```diff
@@ -0,0 +1,37 @@
...
+        patterns: [ "*" ]
+
+  - package-ecosystem: "uv"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "saturday"
+      time: "18:00"
+    groups:
+      production:
```
ojeda-e, line 22: splitting `production` to have main deps updated separately from `development` deps.

.github/dependabot.yml
```diff
@@ -0,0 +1,37 @@
...
+        patterns: ["*"]
+        dependency-type: "production"
+      development:
+        patterns: ["*"]
```
ojeda-e, line 26: This line is to group all bumps/updates in a single PR rather than 1 PR per bump.

ojeda-e (commented):
Some comments below to clarify the dependabot cfg file.

ojeda-e:
@tylerjereddy I intentionally didn't bump micromamba (I did the change and it passes with the bump v1 -> v3) to check that after merge dependabot catchs it and bumps it as I expect. 
I was careful about the format for the version that dependabot reads as well.

.github/dependabot.yml
```diff
@@ -0,0 +1,37 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
tylerjereddy, line 6: I do suggest using a "cooldown" period on new releases like i.e., NumPy does here: https://github.com/numpy/numpy/blob/main/.github/dependabot.yml#L8

It is a protection mechanism against supply chain problems -- they are far more likely to be discovered and fixed if you wait a week or two before allowing the update of a released package (separate from the interval for updates, which is just how often it is allowed to open PRs).

This came up in a recent grant call with sklearn/NumPy teams...

tylerjereddy:
I should also mention -- the rest of the diff looks quite good to me, beyond my one comment. I admit that I didn't spot check all of your hashes for correctness. I need to help Oli with some grant stuff this weekend and SciPy conference/ASCR conference are coming up, so a bit rushed..

ojeda-e:
Thanks for your input @tylerjereddy.
I've added the cooldown as suggested. I do appreciate the time you take to check my PRs! :)

PR #228, opened 2026-07-11
Apply FFT filter on a per-frame basis

### LINKED ISSUE #227: Bug: FFT filtering is applied per surface instead of per-frame
With `surface_method='binning'` and `fft_filter` enabled, the brick-wall filter was applied only to the time-`averaged z_surface` in `_conclude()`, and the mean/Gaussian curvatures were computed from that single filtered average.

Per-frame `results.z_surface`, `results.mean`, and `results.gaussian` were left unfiltered, which computes the wrong value for the estimator because curvature is nonlinear, so `curvature(mean(z)) != mean(curvature(z))`.
Then, when running MC with binning + filter vs. Fourier, they were displaying very different patterns. The filter should be applied in a way that computes curvature per frame and averages afterward, to be consistent with Fourier.

Another issue is that the filter is too aggressive. The brickwall filter also amplifies the second derivative (affecting K). A better filter should be implemented.

## Expected Behaviour
- Filter each per-frame binned surface before computing curvature.
- Report average maps as nanmean of the per-frame quantities (curvature-then-average), matching the Fourier method.
 - Use a smoother filter instead of a brick-wall mask to limit Gibbs ringing, which second-derivative curvature amplifies.

ojeda-e:
Fixes #227 

## Description

- Apply filter brick-wall filter to each frame before curvature is computed + updated tests and docstrings. 

## PR Checklist

 - [x] Tests updated/added?
 - [ ] Documentation updated/added?

### DIFF
--- a/membrane_curvature/base.py
+++ b/membrane_curvature/base.py
@@ -95,15 +95,14 @@ class MembraneCurvature(AnalysisBase):
         Brick-wall filter on the binned height field when ``surface_method='binning'``.
         Default ``'auto'`` (low-pass ``(0, 0.5 * q_Nyq)`` from bin widths). Pass
         ``None`` to disable. Pass ``{'q': (q_low, q_high)}`` for custom bounds in rad/Å.
-        For **average** maps: time-average of ``z_surface``, filter once, then curvature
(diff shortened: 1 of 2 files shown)


ojeda-e:
Closing since #227 is also closed

PR #240, opened 2026-07-16
Enhance: Speed up binning with masked `np.add.at` accumulation

### LINKED ISSUE #225: Investigate binning: should it scale better and run faster?
On small fixtures like the ones used in #200, or in tests, curvature analysis is already fast (~0.6 ms/frame according to the  benchmark results). However, I have been working on large systems (5 microsec simulation time) using MC and I noticed the binning method takes longer than I expected. This problem shows up only when scale increases. This makes sense, because cost scales ~ `n_frames × n_atoms`. A 5 microsec trajectory with thousands of frames on a large bilayer runs a bit slow, at least from what I expect from binning.

I am wondering if, with a fixed grid size, 2x the number of reference atoms roughly runs in 2x runtime is an indication that binning accumulation is contributing to the runtime. If runtime scales linearly with reference atom count at fixed grid size, per-atom binning accumulation is likely a major cost, then profiling `get_z_surface` vs the rest of the binning pipeline will confirm this before implementing a solution.

If the per-atom calculation is a significant contributor to binning cost, a likely candidate is the accumulation loop in get_z_surface:

https://github.com/MDAnalysis/membrane-curvature/blob/8b59e9eab098dd0c7c94bae06b4c213a347b4821/membrane_curvature/binning_surface.py#L145-L150

Even though `surface_method='fourier'` is the default, and in the benchmarks binning is faster than fourier (expected), an improvement for the binning method is still worth it.

ojeda-e:
Fixes #225

## Description

Faster binning surface using a masked grid and accumulting with `np.add.at`.

- Built mask for valid grid -> replaces try/except
- Accumulate with `np.add.at` instead of running Python loop.
- Replace `WarnOnce` with a out of grid warning for total count.

## PR Checklist

 - [N/A] Tests updated/added?

### DIFF
--- a/membrane_curvature/binning_surface.py
+++ b/membrane_curvature/binning_surface.py
@@ -29,30 +29,6 @@ Functions
 
 import numpy as np
 import warnings
-import logging
-
-logger = logging.getLogger('MDAnalysis.MDAKit.membrane_curvature')
-
-
-class WarnOnce:
-    """
-    Class to warn atoms out of grid boundaries only once with full message.
-    After the first occurrence, message will be generic.
-    """
-
-    def __init__(self, msg, msg_multiple) -> None:
-        self.msg = msg
-        self.warned = False
-        self.warned_multiple = False
-        self.msg_multiple = msg_multiple
-
-    def warn(self, *args):
-        if not self.warned:
-            self.warned = True
-            warnings.warn(self.msg.format(*args))
-            logger.warning(self.msg.format(*args))
-        elif not self.warned_multiple:
-            warnings.warn(self.msg_multiple)
 
 
 def derive_surface(atoms, n_cells_x, n_cells_y, max_width_x, max_width_y):
@@ -84,24 +60,6 @@ def derive_surface(atoms, n_cells_x, n_cells_y, max_width_x, max_width_y):
     return get_z_surface(
         coordinates, n_x_bins=n_cells_x, n_y_bins=n_cells_y, x_range=(0, max_width_x), y_range=(0, max_width_y)
     )
-
-
-# messages for warnings, negative and positive coordinates.
-msg_exceeds = 'More than one atom exceed boundaries of grid.'
-negative_coord_warning = WarnOnce(
-    'Atom with negative coordinates falls '
-    'outside grid boundaries. Element '
-    "({},{}) in grid can't be assigned."
-    ' Skipping atom.',
-    msg_exceeds,
-)
-positive_coord_warning = WarnOnce(
-    'Atom coordinates exceed size of grid '
-    "and element ({},{}) can't be assigned. "
-    'Maximum (x,y) coordinates must be < ({}, {}). '
-    'Skipping atom.',
-    msg_exceeds,
-)
 
 
 def get_z_surface(coordinates, n_x_bins=10, n_y_bins=10, x_range=(0, 100), y_range=(0, 100)):
@@ -146,11 +104,8 @@ def get_z_surface(coordinates, n_x_bins=10, n_y_bins=10, x_range=(0, 100), y_ran
     np.add.at(grid_norm_unit, (cell_x_floor[valid_mask], cell_y_floor[valid_mask]), 1)
 
     if not np.all(valid_mask):
-        for index_l, index_m in zip(cell_x_floor[~valid_mask], cell_y_floor[~valid_mask]):
-            if index_l < 0 or index_m < 0:
-                negative_coord_warning.warn(index_l, index_m)
-            else:
-                positive_coord_warning.warn(index_l, index_m, x_range[1], y_range[1])
+        atoms_outside_grid = np.count_nonzero(~valid_mask)
+        warnings.warn(f'{atoms_outside_grid} atoms fall outside the grid boundaries. Skipping atoms.')
 
     z_surface = normalized_grid(grid_z_coordinates, grid_norm_unit)
 
--- a/membrane_curvature/binning_surface.py
+++ b/membrane_curvature/binning_surface.py
@@ -140,19 +140,18 @@ def get_z_surface(coordinates, n_x_bins=10, n_y_bins=10, x_range=(0, 100), y_ran
     cell_x_floor = np.floor(x_coords * x_factor).astype(int)
     cell_y_floor = np.floor(y_coords * y_factor).astype(int)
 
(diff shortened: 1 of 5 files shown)


membrane_curvature/binning_surface.py
```diff
@@ -140,19 +140,17 @@ def get_z_surface(coordinates, n_x_bins=10, n_y_bins=10, x_range=(0, 100), y_ran
...
-        try:
-            # negative coordinates
-            if index_l < 0 or index_m < 0:
-                negative_coord_warning.warn(index_l, index_m)
-                continue
+    valid_mask = (cell_x_floor >= 0) & (cell_y_floor >= 0) & (cell_x_floor < n_x_bins) & (cell_y_floor < n_y_bins)
 
-            grid_z_coordinates[index_l, index_m] += index_z
-            grid_norm_unit[index_l, index_m] += 1
+    np.add.at(grid_z_coordinates, (cell_x_floor[valid_mask], cell_y_floor[valid_mask]), z_coords[valid_mask])
```
ojeda-e, line 103: It's the first time I'm implementing [np.ufunc.at](https://numpy.org/doc/stable/reference/generated/numpy.ufunc.at.html), @orbeckst do you have any comments or objections? Maybe there's even a better way for this.

ojeda-e (commented):
Ran a quick benchmark for:

system a) A membrane-protein system with ~900 atoms of reference and ~2500 frames:

- **This branch: 1.04 ms/frame**
- Main branch: 1.40 ms/frame
- Performance improvement: ~26% reduction in runtime (1.35x faster)

system b) A membrane-only system with ~800 atoms of reference and ~2500 frames:

- **This branch: 0.95 ms/frame**
- Main branch: 1.36 ms/frame ms/frame
- Performance improvement: ~30% reduction in runtime (1.43x faster)

and I'm happy with this.

membrane_curvature/binning_surface.py
tylerjereddy: It is an in place operation so probably pretty efficient already. IMO the speedups described wouldn't justify a large increase in code complexity, but new version doesn't seem too bad.

Your "warning" codepath does look more complex though--I wonder if, in some cases where you hit the warnings, your performance degrades a bit. If the warnings are relatively rare in practice it may be "ok."

ojeda-e: > our "warning" codepath does look more complex though--I wonder if, in some cases where you hit the warnings, your performance degrades a bit. If the warnings are relatively rare in practice it may be "ok."

Yes, this is correct. That path is more complex and it can happen mostly in 2 cases: when MC runs w/o coordinate wrapping and atoms fall outside the box, or in membrane-protein systems that require postprocessing and MC should run with `wrap=False`(There are some other cases but they are kind of forced in the API and not the standard practice).

For the first case the warnings are rare. For the second one they can be frrequent -depending on the system, the number of frames, etc. But anyway, they can be signficant. 

(Ouch, but) thanks for pointing this out. I think I can add those change here as well.

membrane_curvature/binning_surface.py
```diff
@@ -29,30 +29,6 @@
 
 import numpy as np
 import warnings
-import logging
```
ojeda-e, line 32: I ended up removing `WarnOnce` and leftovers from the logger. I hesitated a bit here because I would have preferred to make the warnings somehow resusable. But I think that once we have several atoms, the detailed warning is not meaningful. In the end the aim here is to let the user know some atoms are outside the grid and the calculation skips those.

membrane_curvature/binning_surface.py
```diff
@@ -140,19 +98,14 @@ def get_z_surface(coordinates, n_x_bins=10, n_y_bins=10, x_range=(0, 100), y_ran
...
-            grid_norm_unit[index_l, index_m] += 1
+    np.add.at(grid_z_coordinates, (cell_x_floor[valid_mask], cell_y_floor[valid_mask]), z_coords[valid_mask])
+    np.add.at(grid_norm_unit, (cell_x_floor[valid_mask], cell_y_floor[valid_mask]), 1)
 
-        # too large positive coordinates
-        except IndexError:
-            positive_coord_warning.warn(index_l, index_m, x_range[1], y_range[1])
+    if not np.all(valid_mask):
+        atoms_outside_grid = np.count_nonzero(~valid_mask)
+        warnings.warn(f'{atoms_outside_grid} atoms fall outside the grid boundaries. Skipping atoms.')
```
ojeda-e, line 108: Since there is only one warning, I just left it as `warnings.warn` w/o reusing the WarnOnce messages from before. Yes, I am losing the first `(index_l, index_m)`, but in the end I doubt it provides much value. I'm incline to think the user would care more about the total number of atoms falling outside the grid, particularly in posprocessed trajs for membrane-protein systems.

I'm open to consider other options.

membrane_curvature/tests/test_membrane_curvature.py
```diff
@@ -1299,7 +1299,7 @@ def test_test_analysis_no_wrapping(self, universe):
     def test_positive_coordinates_exceed_grid(self, x_bin, y_bin, box_dim, dummy_array):
         u = mda.Universe(dummy_array, n_atoms=len(dummy_array))
         u.dimensions = [box_dim, box_dim, 300, 90.0, 90.0, 90.0]
-        regex = r'exceed boundaries | size of grid'
+        regex = r'atoms fall outside the grid boundaries'
```
ojeda-e, line 1302: Kept the same unit tests to differentiate exceeding coords from both boundaries (positive and negavtive -below). I think this is better than a single one.

ojeda-e (commented):
@tylerjereddy I added the changes to change the path of the atoms that fall outside the grid. I think there are several ways to implement this, but I went with this for simplicity. Left some comments below explaining my choices.

tylerjereddy:
Apologies if I don't get back to it for a bit (on travel at SciPy '26 conference, then more work travel Sunday), but don't let me block you.

From a *very* quick look at the latest diff, it looks like the warning path is now also "vectorized" so there seems to be not much left to complain about.

ojeda-e:
> Apologies if I don't get back to it for a bit (on travel at SciPy '26 conference, then more work travel Sunday), but don't let me block you.
> 
> From a _very_ quick look at the latest diff, it looks like the warning path is now also "vectorized" so there seems to be not much left to complain about.

Thank you! Your comments are always welcome in the project. Happy travels and enjoy your conference!

PR #278, opened 2026-08-15
Policy: Add SECURITY policy

### LINKED ISSUE #275: Policy: Add a security policy
Set instructions to report a security vulnerability in the project.. See https://docs.github.com/en/code-security/how-tos/report-and-fix-vulnerabilities/configure-vulnerability-reporting/add-security-policy

To fix this issue:
- [x] Add a security policy in `SECURITY.md` file.

ojeda-e:
Fixes #275 

**This sits until the 2.0.0 release.**

## Description

Added security policy with supported versions. This is a missing item in the [recommended Github community standards](https://github.com/MDAnalysis/membrane-curvature/community).

## LLM / AI acknowledgement and disclosure

- [x] I have read and understood the [MDAnalysis AI Policy](https://github.com/MDAnalysis/mdanalysis/blob/develop/AI_POLICY.md) that governs this project.
- LLMs or other AI-powered tools (beyond simple IDE use cases) were used in this contribution: no

## PR Checklist

 - [N/A] Tests updated/added?
 - [N/A] LLM/AI disclosure was updated?

## Do not merge until 2.0.0 is released

This PR is intentionally **blocked from merging** until the `2.0.0` release is tagged and published.

The Supported Versions section is written against `2.0.0` as the supported line. Merging before `2.0.0` exists would ship a policy that points users to a release they cannot install yet.

### Merge checklist
- [ ] `v2.0.0` is tagged and published.

Marking as **draft** to enforce the block. Marking to "Ready for review" once 2.0.0 ships.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,37 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+Please report security vulnerabilities by opening a new [GitHub security
+advisory](https://github.com/MDAnalysis/membrane-curvature/security/advisories/new).
+
+> [!WARNING]
+> Please do **not** file a public GitHub issue for security reports.
+
+When reporting, if possible, include:
+- A clear description of the issue and potential impact.
+- Steps to reproduce or a proof of concept.
+- Affected MembraneCurvature version and environment details.
+- Any suggested mitigations.
+
+You can also send an email to `mdanalysis@numfocus.org`, which is an alias to
+a subset of the MDAnalysis maintainers' team.
+
+If the security vulnerability is accepted, a patch will be crafted privately
+in order to prepare a dedicated bugfix release as timely as possible (depending
+on the complexity of the fix).
+
+## Supported Versions
+
+Only the main branch (the 2.x release line) receives updates, including
+security fixes. Older release lines (e.g., 1.1.x) are not maintained. If you are
+using an older version, please upgrade to the latest 2.x release to receive
+fixes.
+
+When reporting a potential vulnerability, confirm that it reproduces against
+the latest 2.x version.
+
+| Version       | Supported          |
+| ------------- | ------------------ |
+| 2.0.0         | ✅ |
+| < 1.1.2       | ❌ |
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -23,15 +23,17 @@ on the complexity of the fix).
 
 ## Supported Versions
 
-Only the main branch (the 2.x release line) receives updates, including
-security fixes. Older release lines (e.g., 1.1.x) are not maintained. If you are
-using an older version, please upgrade to the latest 2.x release to receive
-fixes.
+**Security fixes are applied only to the latest release.** Older releases do not
+receive fixes. If you are using an older version of MembraneCurvature, please
+upgrade to the latest release to receive security updates.
+
+> [!NOTE]
+> The current supported release line is 2.x (latest: 2.0.0).
 
 When reporting a potential vulnerability, confirm that it reproduces against
-the latest 2.x version.
+the latest release.
 
 | Version       | Supported          |
 | ------------- | ------------------ |
-| 2.0.0         | ✅ |
+| Latest 2.x release | ✅ |
 | < 1.1.2       | ❌ |


SECURITY.md
```diff
@@ -0,0 +1,37 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+Please report security vulnerabilities by opening a new [GitHub security
+advisory](https://github.com/MDAnalysis/membrane-curvature/security/advisories/new).
```
ojeda-e, line 6: I leveraged the private vulnerability reporting that is enabled in the project here.

SECURITY.md
```diff
@@ -0,0 +1,37 @@
...
+> [!WARNING]
+> Please do **not** file a public GitHub issue for security reports.
+
+When reporting, if possible, include:
+- A clear description of the issue and potential impact.
+- Steps to reproduce or a proof of concept.
+- Affected MembraneCurvature version and environment details.
+- Any suggested mitigations.
+
+You can also send an email to `mdanalysis@numfocus.org`, which is an alias to
```
ojeda-e, line 17: Just want to be sure adding this email is okay. @orbeckst I'm tagging you here since you are the main policy reviewer. Maybe coredevs is better? Not sure, I saw this email is the one in [mdadash](https://github.com/search?q=org%3AMDAnalysis+SECURITY.md&type=code)

SECURITY.md
```diff
@@ -0,0 +1,37 @@
...
+a subset of the MDAnalysis maintainers' team.
+
+If the security vulnerability is accepted, a patch will be crafted privately
+in order to prepare a dedicated bugfix release as timely as possible (depending
+on the complexity of the fix).
+
+## Supported Versions
+
+Only the main branch (the 2.x release line) receives updates, including
```
ojeda-e: My main point to convey here is that I will support >2.x because most of 1.x is with versions of Python that unsupported as well.

ojeda-e:
I am adding @orbeckst and @tylerjereddy as reviewers. There is no rush at all, it will sit here until next release. Unless it's better to add the policy with the current versions now and change it to 2.0.0 after release, which is also an option.
Thanks!

SECURITY.md
tylerjereddy: Yeah, that looks right. Some larger projects request security reporting through https://www.sonarsource.com/solutions/security/?tidelift=tidelift-redirect, but this seems correct here. If you do ever receive one, note that it can take weeks/months to get the GitHub folks to properly review it.

It also isn't fully clear what the correct workflow is with respect to doing releases that patch CVEs--i.e., do you need to rush it out the door quietly, not describe the patch in detail in the commit message on release branch? I brought this up at SciPy 2026 conference re: publishing some better guidelines on that.

SECURITY.md
tylerjereddy: I think for SciPy we setup a separate vulnerability email list because the discussions get long and complex and the number of people that are copied is too large for that, but this should be quite rare to happen for MDAnalysis I suspect (even with the surge of LLM exposed vulnerabilities).

SECURITY.md
```diff
@@ -0,0 +1,37 @@
...
+- Steps to reproduce or a proof of concept.
+- Affected MembraneCurvature version and environment details.
+- Any suggested mitigations.
+
+You can also send an email to `mdanalysis@numfocus.org`, which is an alias to
+a subset of the MDAnalysis maintainers' team.
+
+If the security vulnerability is accepted, a patch will be crafted privately
+in order to prepare a dedicated bugfix release as timely as possible (depending
+on the complexity of the fix).
```
tylerjereddy, line 22: Yeah, as noted elsewhere, I wonder if we have to rush as release managers, be subtle about the commit messages, or if we can really just do a pretty standard release at the end of the day. Not sure about others, but usually rushing causing me to make mistakes so yeah, but hopefully we get nice community guidelines on what to do in this regard soon.

SECURITY.md
tylerjereddy: Interesting--you do releases right off the `main` branch? I guess I've been doing it differently for so long I don't even think about alternatives to using maintenance branches.

tylerjereddy (commented):
I added some comments, mostly just rambling about the state of security patching/releases at the moment.

SECURITY.md
ojeda-e: No, I use tags and the deployment workflow in the CI. Thanks for flagging this, I didn't rephrase it correctly.

Maybe what I should say is that Only the _latest_ release from the main branch, which would be the 2.x and not older, will receive fixes. I'll update this.

ojeda-e:
Thanks for your comments @tylerjereddy, after your review I rephrased the section for supported versions.

Longer term it'd be great to have org-level guidelines so each repo isn't figuring this out on its own. Until then, this policy at least gives us a starting point and some pointers to follow.

